PR #7, opened 2019-01-17
correction to `type` data in example

PapillonMcGill:
implementing JB's remark about type.  Left indentation so it is easier to read for a human.


jbpoline:
Sounds right - merging ! please go on and try to implement more of the DATS elements for this example, and for another example (preventAD).

PR #13, opened 2019-04-01
example genomics metadata

emmetaobrien:
This update adds 26 files of metadata for 1000 Genomes Project data provided by David Bujold at C3G, which document variation between 1093 human genome sequences.   One of these is a master file, and 25 contain metadata for individual chromosomes (1-22, X, Y, and the mitochondrial genome).  File structure has been validated against the DATS dataset_schema.json file from the bioCADDIE GitHub repository.  The metadata for each individual chromosome include names, file creation dates, locations for the zipped .vcf file containing the underlying data,  summary statistics of the single-nucleotide polymorphisms and insertion/deletion events recorded in each .vcf file, and  an "extraProperties" entry providing a link to the reference sequence for each chromosome.  The master file includes a primary publication reference, formal taxonomic identification, and an appropriately formatted list of the part files.  Design here principally by EOB, with input from JB and Jennifer at the MONII hackathon.

### DIFF
--- a/metadata/example/1KGP_chr1.json
+++ b/metadata/example/1KGP_chr1.json
@@ -1,6 +1,11 @@
 {
 	"version": "1.0",
 	"privacy": "public open",
+	"licenses": [
+		{
+			"name": "to be determined"
+		}
+	],
 	"identifier": {
 		"identifier": "1KGP_chr1"
 	},
--- a/metadata/example/1KGP_chr1.json
+++ b/metadata/example/1KGP_chr1.json
@@ -1,0 +1,60 @@
+{
+	"version": "1.0",
+	"privacy": "public open",
+	"identifier": {
+		"identifier": "1KGP_chr1"
+	},
+	"title":"Chromosome 1",
+	"dates": [
+		{
+			"type": {
+				"value":"source .vcf file creation date"
+			},
+			"date": "2015-02-18 00:00:00"
+		}
+	],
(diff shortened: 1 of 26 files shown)


paiva:
Great work @emmetaobrien! We will review it soon!

jbpoline:
looks great - congrats ! some  quick questions / remarks: 
- could the scripts that generated the files be committed ? (if this was scripted) 
- would be interesting to see what the schema.org file would be for this
- would be cool to see if the datalad dataset could be created for these data. I guess at some stage we may want to have an extractor to ingest this into the datalad meta data - wdyt ?

glatard:
Hi, nice work! Is this a datalad dataset? I don't see the links to the data being described by the metadata. If not, is there a plan to add the (links to the) data in the future?

emmetaobrien:
jbpoline: I can certainly commit the two scripts used for generating these datasets and the one used for validating them, where in the repository would be a sensible place to put them ?

glatard: URLs to the data are in 'storedIn" attributes in the file for each chromosome. I had not been planning to include the data directly in his repository.

glatard:
With datalad you can "declare" the files to the repo without adding their content. It allows to associate the metadata with the actual files. In case anyone wants to retrieve them, conditional to permissions of course, they can just `datalad get` them. The process of adding data is explained at https://github.com/CONP-PCNO/conp-dataset.

PR #19, opened 2019-04-09
improvement to SDas metadata file

PapillonMcGill:
added dimension
added few other details

### DIFF
--- a/metadata/example/SDasDataset.json
+++ b/metadata/example/SDasDataset.json
@@ -1,6 +1,7 @@
 {
     "schema": "https://github.com/biocaddie/DATS/blob/master/dataset_schema.json",
-    "title": "Humain MRI Phantom",
+    "title": "S. Das Human MRI Phantom ",
+    "description": "Human Brain phantom scans, T1 weighted images and others on n=? sites. The data have been collected on a single participant and are available in minc format. ",
     "description": "Multiple MRI scans of a single human phantom over 11 years",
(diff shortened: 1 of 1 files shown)


glatard:
Could the files be added to the dataset too?

PapillonMcGill:
It will be.

metadata/example/SDasDataset.json
```diff
@@ -1,17 +1,35 @@
 {
     "schema": "https://github.com/biocaddie/DATS/blob/master/dataset_schema.json",
-    "title": "SDas Phantom",
-    "description": "Samir Das phantom scans",
+    "title": "Humain MRI Phantom",
```
paiva: `Human` instead of `Humain`. A better title would be `Multicenter Human MRI Phantom`, the multicenter part is what makes this dataset attractive.

paiva:
I think it would good to add the source of the dataset: `https://phantom-dev.loris.ca/`

metadata/example/SDasDataset.json
jbpoline: Agreed, something  like

```suggestion
    "title": "Human MRI Phantom",
    "description": "Samir Das phantom scans, T1 weighted images on many different sites",
```

Also,  if we do searches on the "description" field it would be good to put relevant keywords there - for instance I think this could be considered as a "longitudinal" study and if we know the other type of images we could add them in the description as well

PR #30, opened 2019-04-21
[Dataset] Add samir phantom dataset

paiva:
This PR:
* Adds Samir's multicenter phantom dataset, using the workflow from `conp-dataset`


jbpoline:
datalad install yields correct urls - merging

PR #33, opened 2019-04-24
[Dataset] Add PREVENT-AD Open

paiva:


paiva:
@PapillonMcGill how about now?

PapillonMcGill:
@paiva Have you done a commit? I see no change.

paiva:
@PapillonMcGill can you try now? It should be consistent

jbpoline:
Just checked with @paiva  that this urls are fine and you can download files when authenticated in loris, merging. 

Still : DATS model should be in the dataset

PR #49, opened 2019-06-13
add submodule, cool

dlq:
Enhanced Points of Interest for Montreal for 2018.  No metadata.


paiva:
@dlq I forgot to ask, is this a real PR or just to test it out? If it is real, I can start testing it right away

dlq:
Real.  I will update it to be even more real with metadata, etc as I go. 


Darcy

paiva:
Hi @dlq, this PR is good and works well, but we now require that datasets include a dats.json file with metadata about the dataset. We can follow up on this anytime this week.

dlq:
I will resubmit this PR when it’s actually ready. I shouldn’t have let it sit for so long.

Issue #16, opened 2019-04-08
Continuous integration of datasets


paiva:
* We need to create a document outlining how to test PRs and to decide what is a good dataset to merge on `conp-dataset`

* It would be nice to build some automation around this, maybe look into Travis to see if we can automate some of the tests (like valid links)

* Also, we will require 2 reviewers before merging a PR

paiva:
Should this be a document inside this repository or part of `conp-documentation`?

glatard:
I think it should be a markdown file in this repo and a `.travis.yml` file.

jbpoline:
I'll create something on the doc

glatard:
Sorry @jbpoline, I really think these requirements should be in conp-dataset as it is very specific to this repo and might evolve with it.

glatard:
We should also consider creating a `develop` branch for the dataset and have release cycles.

glatard:
Can we list here the requirements before a PR can be merged, before we add them to Travis tests? 
[ ] dataset can be installed, recursively if it has sub-datasets
[ ] every file in the dataset has a URL
[ ] every file can be retrieved *or* requires authentication (not sure how to test when authentication is required)
[ ] DATs model is present and valid

jbpoline:
Shouldn't we also testing that when logged on the portal we can actually download the data ? Not sure that can be automated but we shouldn't have hundreds of datasets soon, so to start with this could be manual.

glatard:
Here is a summary of TODOs regarding this:
- Create a .travis.yml file and add the project to Travis. Add Travis badge to README.md
- Using Datalad's Python API, create a test script that iterates through the sub-datasets and checks that:
    * Dataset can be installed, recursively if it has sub-datasets
    * Every file in the dataset has a URL
    * Every file can be retrieved or requires authentication (not sure how to test when authentication is required)
    * DATs model is present and valid
    * If dataset is derived data, raw data is a sub-dataset
- Add a PR template containing this checklist.

Notes:
* It might be too heavy to test "every file in the dataset" in every test build. I guess a random sub-sample of, say, 20 files max should be fine.
* [Boutiques](https://github.com/boutiques) contains an example of .travis.yml and PR template (in dev branch).
* @jbpoline testing that the datasets can be downloaded from the portal should be tested in the portal repo, not here. Here we should only make sure that the DataLad repo is clean and working.
[...]

PR #55, opened 2019-07-02
Add test script and integration with Travis.CI

joeyzhou98:
Test script checks for:
- Dataset can be installed, recursively if it has sub-datasets
- Every file in the dataset has a URL
- Every file can be retrieved or requires authentication
- DATs model is present and valid
- If dataset is derived data, raw data is a sub-dataset

Issue #16


joeyzhou98:
Hi @glatard let me know if there is anything else to change

paiva:
@joeyzhou98 Can you resolve merge conflicts? Thanks

joeyzhou98:
> @joeyzhou98 Can you resolve merge conflicts? Thanks

@paiva all yours

tests/create_tests.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+
+dataset_titles = list(filter(lambda x: x[0] != ".", os.listdir("projects") + os.listdir("investigators")))
+
+for title in dataset_titles:
+    with open("tests/test_" + title + ".py", "w") as f:
+        f.write("from functions import examine\n\n\n")
```
glatard: use `os.linesep` instead of `\n`

tests/create_tests.py
```diff
@@ -0,0 +1,10 @@
...
+        f.write("def test_" + title.replace("-", "_") + "():\n")
+        f.write("\tassert examine('" + title + "') == 'All good'\n")
```
glatard: Replace by template containing a key to be substituted by `title` before writing.

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+def examine(dataset):
+
+    # Number of files to test in each dataset
+    num_files = 100
+
+    # Root directory can be projects or investigators
+    root_dir = "projects" if dataset in listdir("projects") else "investigators"
+
+    # Count the number of testable files while ignoring files in directories starting with "."
+    count = sum([len(files) if r.split("/")[-1][0] != "." else 0 for r, d, files in walk(root_dir + "/" + dataset)])
```
glatard: * use `os.path.join` instead of the 2nd `/`.
* use `os.path.basename(os.path.dirname(r))` instead of `r.split("/")[-1]`

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Calculate the odds to test a file
+    odds = num_files/count
+
+    # Start to test dataset
+    return recurse(abspath(root_dir + "/" + dataset), odds)
```
glatard: use `op.join` instead of `/`

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+    for file in files:
+
+        # Execute any config.sh scripts
+        if file == "config.sh":
+            with open(file, "r") as f:
+                line = f.readline()
+                while line:
+                    subprocess.call(line)
+                    line = f.readline()
+            continue
```
glatard: Replace by a line of bash in `.travis.yml`, executed before the tests are run: something like `find . -name config.sh | xargs bash`.

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # If the file starts with ".", has descriptor or readme in filename, continue onto next file
+        if file[0] == "." or \
+            "descriptor" in file.lower() or \
+            "readme" in file.lower():
```
glatard: Put lines between '( )' instead of using `\`.

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+import subprocess
+
+
+def recurse(directory, odds):
+
+    # Get all file names in directory
+    files = listdir(directory)
+
+    # Loop throw every file
+    for file in files:
```
glatard: Rename `file` to `file_name`

tests/functions.py
glatard: You should only exclude `.git` and `.datalad`.

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+            else:
+                return "No annexkey in file: " + full_path
+
+            # Check for authentication
+            if msg["status"] == "error" and \
+                "unable to access" not in msg["message"].lower() and \
+                "not available" not in msg["message"].lower():
+                return "No authentication setup for file: " + full_path
+
+    return "All good"
```
glatard: Return "All good" if file requires authentication.

tests/functions.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def examine(dataset):
```
glatard: Check if dataset should be installed recursively (most likely yes).

glatard: Check that `dats.json` and `README.md` are present in each dataset.

paiva:
@joeyzhou98 Let me know when it is ready for review. Thanks!

tests/create_tests.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+dataset_titles = list(filter(lambda x: x[0] != ".", listdir("projects") + listdir("investigators")))
+
+template = Template("""from functions import examine
+
+
+def test_$title_no_hyphen():
+    assert examine('$title') == 'All good'
+""")
```
glatard: great finding of the Template class!

tests/functions.py
```diff
@@ -0,0 +1,72 @@
...
+from random import random
+import datalad.api as api
+
+
+def recurse(directory, odds):
+
+    # Get all file names in directory
+    files = listdir(directory)
+
+    # Loop throw every file
```
glatard: typo: *through

joeyzhou98:
Hi, @paiva and @glatard, it is ready for another round of review!

tests/functions.py
```diff
@@ -0,0 +1,72 @@
...
+
+        full_path = join(directory, file_name)
+
+        # If the file is a directory
+        if isdir(full_path):
+
+            result = recurse(full_path, odds)
+
+            if result != "All good":
+                return result
```
glatard: I think you could just `return recurse(full_path, odds)`, since you're returning "All good" later anyway.

tests/functions.py
```diff
@@ -0,0 +1,72 @@
+from os import listdir, walk
+from os.path import isdir, exists, join, abspath, basename, dirname
+from random import random
+import datalad.api as api
+
+
+def recurse(directory, odds):
```
glatard: You should add a docstring to explain what the function does and what `odds` is.

tests/functions.py
```diff
@@ -0,0 +1,72 @@
...
+                if "URL" not in msg["annexkey"]:
+                    return "No URL in annexkey: " + msg["annexkey"] + " for file: " + full_path
+            else:
+                return "No annexkey in file: " + full_path
+
+            # Check for authentication
+            if (msg["status"] == "error" and
+                    "unable to access" not in msg["message"].lower() and
+                    "not available" not in msg["message"].lower()):
+                return "No authentication setup for file: " + full_path
```
glatard: 1. Here we should return "Cannot download file and didn't hit authentication request".
2. Why is "not available" not an error?

joeyzhou98: Indeed, I will remove the check for "not available"

joeyzhou98:
@glatard I wasn't able to find a way to run the tests non-interactively for some projects, but it seems like on Travis, it wasn't a problem: https://travis-ci.com/joeyzhou98/conp-dataset/builds/119529868; this build ran the tests when the portion of the code checking for dats.json and README.md was commented out, simulating that those files are already there

image

and Travis didn't seem to have hit the interactive authentication. My guess is that Travis has some way of bypassing interactiveness. I will have to do further tests

tests/functions.py
```diff
@@ -50,27 +38,35 @@ def recurse(directory, odds):
...
 
-    # Number of files to test in each dataset
-    num_files = 100
-
     # Root directory can be projects or investigators
     root_dir = "projects" if dataset in listdir("projects") else "investigators"
+    full_dir = join(root_dir, dataset)
+
+    # Check if dats.json and README.md are present in root of dataset
+    # if "dats.json" not in listdir(full_dir):
```
gi114: Hi @joeyzhou98 do you validate at all whether the dats model file is a a valid json-ld or if there are errors, apart from checking if the file exists? I have some work on that that we can add in case. Thanks

glatard: This is out of the scope of this current PR. As discussed with @PapillonMcGill, this will be added to Travis once this PR is merged.

.gitignore
```diff
@@ -0,0 +1,4 @@
+.idea/
```
paiva: Could you add `.DS_Store as` well?

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+
+python:
+- 3.6
+
+before_install:
+- bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)
+- sudo apt-get -q install -y git-annex-standalone
+
+install:
+- pip install datalad
```
paiva: Does it make a difference if you specify `pip3` instead of `pip`? I am not sure if it is irrelevant to say pip3 given that you are specifying Python 3.6 above. Just wondering...

glatard: It shouldn't make any difference when Python 3 is used.

joeyzhou98: I just tested and it seems like it doesn't make a difference with `pip3`: https://travis-ci.com/joeyzhou98/conp-dataset/jobs/218328117 compared to a previous build with only `pip`

.gitignore
joeyzhou98: Done

.travis.yml
paiva: Super, thanks!!

tests/functions.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def examine(dataset):
+
+    # Root directory can be projects or investigators
+    root_dir = "projects" if dataset in listdir("projects") else "investigators"
+    full_dir = join(root_dir, dataset)
+
+    # Check if dats.json and README.md are present in root of dataset
+    if "dats.json" not in listdir(full_dir):
```
paiva: Check for `dats.json` or `DATS.json` (could be uppercase...)

joeyzhou98:
Hi, @glatard the discussed changes are now implemented, let me know if there is anything else!

glatard:
All good to me now! @paiva, would you like to have a second look?

PR #56, opened 2019-07-03
Documentation fixes and suggestions

kaitj:
Made a few changes to the documentation, as well as some suggestions based on the CONTRIBUTING document and from a previous attempt at adding a dataset (might be better fitted to create issues for them?). 

## Changes
- Added a missing word to instructions under "Contribution workflow" heading
- Fix spacing under "Adding meta-data" heading
- Update URL to `metadata/example`
  - Previous URL did not work - not sure if this was intent until metadata is formalized

## Suggestions
- Intended audience for the document, and what is there experience?
  - Maybe move the suggestion about reading GitHub documentation prior to the workflow heading (since there appears to be an underlying assumption of knowing how to use git. 
  - Where should a user go / who to contact if there is an issue - post under issues in the repo? Would have a help or contact section near the top for the above notes
- Instructions for "Adding Data"
  - There seems to be a few different instructions regarding adding a sub-module in different documentations. Again, this could be due to the underlying assumption of git. Had some issues when I was originally setting up the sub-module for the Khanlab datasets. Think it would be good if the documents had the same instructions for the sub-modules.
  - Is there documentation available from potentially different sources (eg. aws, osf, etc.) that be linked to? Depending on the target audience, maybe there should be a note that the files need to be hosted somewhere (unless there is available space for this). 
- Current instructions can also appear to imply a binary workflow - either it is already hosted somewhere (https) or to host through Google Drive, which can be potentially problematic. Suggestion would be to link to a folder of documentations that help walk the user in adding their data using their source. This does not need to be how to host the files, but just how to add from the source (eg. aws, osf, etc.) Additionally, this might help to provide some clearer instructions for users to contribute their dataset.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contribution			
 
-New contributions are more than welcome! ❤️ 
+New contributions are more than welcome! ❤️
 
 ## Contribution workflow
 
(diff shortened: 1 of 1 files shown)


kaitj:
@glatard - Forgot to tag in original PR. Some fixes and suggestions for documentation

PR #75, opened 2019-08-30
Updating SIMON dataset submodule on the CONP-PCNO end

gi114:
## Description
PR for adding dats.json, README.md and gitignore to SIMON dataset

Mandatory files and elements:
- A `README.md` file, at the root of the dataset
- A `dats.json` file, at the root of the dataset
- If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset
- A DOI (see instructions in [contribution guide](https://github.com/CONP-PCNO/conp-dataset/blob/master/.github/CONTRIBUTING.md), and corresponding badge in `README.md`

Functional checks:
- Dataset can be installed using DataLad, recursively if it has sub-datasets
- Every data file has a URL
- Every data file can be retrieved or requires authentication
- `dats.json` is a valid DATs model
- If dataset is derived data, raw data is a sub-dataset


gi114:
Hi @shots47s 

That's great. Thanks for your comment

PR #83, opened 2019-09-13
Preventad readme

PapillonMcGill:
## Description
update to preventad-open to add a README.md file

## Checklist
Mandatory files and elements:
- [] A `README.md` file, at the root of the dataset
- [ ] A `dats.json` file, at the root of the dataset

Functional checks:
- [ ] `dats.json` is a valid DATs model


dlq:
The description of this PR is that it adds a `README.md` but all I’m seeing are multiple changes to the `dats.json` which seem to validate.

Also, I’m not sure what to do with the sub-dataset checks failing in CI.

PapillonMcGill:
not the good one, sorry, will resubmit later

PR #87, opened 2019-09-18
[dataset] added DATS to khanlab datasets

Park-Patrick:
## Description
An update to existing dataset to include metadata.

## Checklist

Mandatory files and elements:
- [x] A `dats.json` file, at the root of the dataset
- [x] If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset

Functional checks:
- [x] `dats.json` is a valid DATs model (as far as we know)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,5 +35,5 @@
   path = projects/nhpatlas
   url = https://github.com/emmetaobrien/nhpatlas.git
 [submodule "investigators/Khanlab/BigBrainMRICoreg"]
-	path = investigators/Khanlab/BigBrainMRICoreg
-	url = https://github.com/khanlab-datasets/BigBrainMRICoreg
+  path = investigators/Khanlab/BigBrainMRICoreg
+  url = https://github.com/khanlab-datasets/BigBrainMRICoreg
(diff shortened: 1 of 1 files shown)


Park-Patrick:
hi @glatard , could you review this PR for us, so we can merge?

joeyzhou98:
> Hi @Park-Patrick, great work!
> 
> The tests are still failing though:
> 
> ```
> Dataset investigators/Khanlab doesn't contain dats.json in its root directory
> ```
> 
> This is because they see `Khanlab` as a dataset and not as a collection of subdatasets. Do you think you could update the tests so that they check sub-datasets instead of the parent? The code is here (see how it's called in `.travis.yml`):
> 
> https://github.com/CONP-PCNO/conp-dataset/blob/b784d8d6e6b73fec7490d015bdc9ac5837748ddf/tests/create_tests.py#L5
> 
> Instead of listing the directories, I guess it would be better to read the submodule file. @joeyzhou98 could help with that if needed. Once it's done, you could check that `pytest` works in your local host.

@glatard changing the tests to look at submodules instead is a big change to tests and might break other people's datasets so are you sure this is what we want?

glatard:
@joeyzhou98 I understand that, but I do think we want to check datasets (sub-modules), not directories. Do you see any other solution to test datasets that have sub-datasets?

joeyzhou98:
@glatard sounds good I can make the changes once I have time. Is it possible to create an issue for it so that we can reference back to it? @Park-Patrick

Park-Patrick:
Related issue #92 created

PR #84, opened 2019-09-13
Perform dats.json

PapillonMcGill:
## Description
adding a dats.json file to PERFORM

## Checklist

Mandatory files and elements:
- [ ] A `README.md` file, at the root of the dataset
- [ ] A `dats.json` file, at the root of the dataset
- [ ] If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset
- [ ] A DOI (see instructions in [contribution guide](https://github.com/CONP-PCNO/conp-dataset/blob/master/.github/CONTRIBUTING.md), and corresponding badge in `README.md`

Functional checks:
- [ ] Dataset can be installed using DataLad, recursively if it has sub-datasets
- [ ] Every data file has a URL
- [ ] Every data file can be retrieved or requires authentication
- [ ] `dats.json` is a valid DATs model
- [ ] If dataset is derived data, raw data is a sub-dataset

PR #88, opened 2019-09-19
[Test function] Timeout download and hanging authentication

mathdugre:
## Description
Currently, some dataset tests are failing due to hanging authentication.
I think that timing out the request should solve that issue.

The timeout prevents two things: (1) hanging prompt for authentication and (2) full download of the data. (1) Prevent tests to fail when credentials are not provided while (2) aims at improving speed.

## Related issues (optional)

This PR should solve the authentication hanging issue with PR #82, #84, #86.

### DIFF
--- a/tests/functions.py
+++ b/tests/functions.py
@@ -55,17 +55,13 @@ def recurse(directory, odds):
 
         # If the file is a broken symlink and with odd chance
         elif not exists(full_path) and random() < odds:
-            try:
-                # Timeouts the download or a hanging authentification
-                with timeout(3): 
-                    msg = api.get(path=full_path, on_failure="ignore", return_type="item-or-list")
-                    
-                    # Check for authentication
-                    if msg["status"] == "error" and "unable to access" not in msg["message"].lower():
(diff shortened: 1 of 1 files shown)


tests/functions.py
```diff
@@ -31,11 +55,17 @@ def recurse(directory, odds):
...
+            try:
+                # Timeouts the download or a hanging authentification
+                with timeout(3): 
+                    msg = api.get(path=full_path, on_failure="ignore", return_type="item-or-list")
+                    
+                    # Check for authentication
+                    if msg["status"] == "error" and "unable to access" not in msg["message"].lower():
+                        return "Cannot download file and didn't hit authentication request for file: " + full_path
+
+            except TimeoutError:
```
dlq: Doesn't the `with timeout(3):` handle the `TimeoutError`?

dlq:
I really don't know what to do with PR's that don't pass the *current* CI checks.

tests/functions.py
mathdugre: Good point, I'll modify that

PR #96, opened 2019-10-03
Update submodules pointing to invalid commits

driusan:
There were 2 submodules pointing to commits that
don't exist, likely because they were overwritten
by a force push. This was causing errors for me
of the form:

```
error: Server does not allow request for unadvertised object 3d6e1454aac5b1ac154c7664dbccaca2e49464ac
Fetched in submodule path 'investigators/Khanlab/BigBrainMRICoreg', but it did not contain 3d6e1454aac5b1ac154c7664dbccaca2e49464ac. Direct fetching of that commit failed.
```

This updates those submodules to point to the hash
of the current tip of their respective repos, so
that "git submodule update --init" works.


xlecours:
@emmetaobrien Is this already fixed?

emmetaobrien:
For the moment CONP-PCNO:master points to my forks of those datasets, so the interface finds them.

driusan:
I guess we should just close this?

Issue #92, opened 2019-09-24
Check datasets not directories


Park-Patrick:
- Needed for cases where datasets have subdatasets 

Related discussion from pull request #87 

> Hi @Park-Patrick, great work!
> 
> The tests are still failing though:
> 
> ```
> Dataset investigators/Khanlab doesn't contain dats.json in its root directory
> ```
> 
> This is because they see `Khanlab` as a dataset and not as a collection of subdatasets. Do you think you could update the tests so that they check sub-datasets instead of the parent? The code is here (see how it's called in `.travis.yml`):
> 
> https://github.com/CONP-PCNO/conp-dataset/blob/b784d8d6e6b73fec7490d015bdc9ac5837748ddf/tests/create_tests.py#L5
> 
> Instead of listing the directories, I guess it would be better to read the submodule file. @joeyzhou98 could help with that if needed. Once it's done, you could check that `pytest` works in your local host.

@joeyzhou98: @glatard changing the tests to look at submodules instead is a big change to tests and might break other people's datasets so are you sure this is what we want?

@glatard: @joeyzhou98 I understand that, but I do think we want to check datasets (sub-modules), not directories. Do you see any other solution to test datasets that have sub-datasets?

@joeyzhou98: @glatard sounds good I can make the changes once I have time. Is it possible to create an issue for it so that we can reference back to it? @Park-Patrick

PR #97, opened 2019-10-08
Make tests check submodules, not directories

joeyzhou98:
## Description
Instead of assuming every directory under "projects" or "investigators" is a dataset, get dataset paths from submodules.path instead from the .gitmodules file. Update tests to reflect the changes.
IMPORTANT: making tests enforce specifically "DATS.json" and "README.md" as written!!!!

## Related issues (optional)
#92

### DIFF
--- a/tests/create_tests.py
+++ b/tests/create_tests.py
@@ -1,17 +1,18 @@
-from os import listdir
 from string import Template
+from git import Repo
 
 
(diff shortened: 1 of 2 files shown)


emmetaobrien:
We agreed at Friday's sprint to standardise on the DATS.json capitalisation pattern, so the test should be modified accordingly.

glatard:
@emmetaobrien, would you like to review this so that we can merge?

PR #105, opened 2019-11-06
Adding DATS.json-ld files

emmetaobrien:
## Description

This PR adds DATS.json-ld files, generated from the DATS.json files, for seven of the CONP datasets (1000GenomesProject, Brainspan, PERFORM, celltypes, mm_neo_atlas, nhpatlas and refseq) and also removes the obsolete descriptor.json file from the PERFORM dataset.  These files have been generated mostly automatically, though some manual intervention is needed in cases of ambiguity (eg, for some of these datasets, the "authors" or "creators" entries are lists of names and count as a People entity in DATS, while for others where a consortium is listed as author that counts as an Organization entity.)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,9 +16,6 @@
 [submodule "projects/SIMON-dataset"]  
   path = projects/SIMON-dataset
   url = https://github.com/conpdatasets/SIMON-dataset
-[submodule "projects/PERFORM"]
-  path = projects/PERFORM
-  url = https://github.com/conpdatasets/PERFORM.git
 [submodule "investigators/Khanlab/HCPUR100-Template"]
   path = investigators/Khanlab/HCPUR100-Template
   url = https://github.com/khanlab-datasets/HCPUR100-Template.git


emmetaobrien:
The code and lookup table used to generate these files is available in https://github.com/neurohub/neurohub_documentation/tree/master/json-to-json_ld

dlq:
Instead of `$schema` I think you mean `@context`.

Your `DATS.json-ld` doesn't check out in [JSON Playground](https://json-ld.org/playground/) 😦 

`.jsonld` is the recommended extension.

dlq:
I see you've made changes but for some reason it's still not checking out in https://json-ld.org/playground/

emmetaobrien:
OK, I think I see what I was getting wrong, and this now looks to me to be behaving appropriately on that site for LD context files referring to schema.org.

dlq:
I hate to be *that* guy but ...  It seems that the json-ld parser is only getting the `organization` and `name` from the entries in the `hasPart` array.

```
  "https://schema.org/hasPart": [
    {
      "https://schema.org/creator": {
        "@id": "https://w3id.org/dats/schema/organization_schema.json",
        "@type": "https://json-ld.org/playground/Organization"
      },
      "https://schema.org/name": {
        "@type": "https://schema.org/Text",
        "@value": "Chromosome 1"
      }
    },
    ...
  ]
```

dlq:
The `DATS.jsonld` has enough information now.

emmetaobrien:
This branch has been updated to remove the PERFORM dataset, to resolve a conflict with incorporating PR #112.

PR #115, opened 2019-11-28
Removed Patrick's and JB's datasets

glatard:
## Description
Removed the following datasets as they seem to be unmaintained tests:
* `investigators/jb` (to be confirmed by @jbpoline)
* `investigators/Park-Patrick` (checked with @Park-Patrick)

## Checklist

No new dataset, checklist isn't relevant.

## Related issues (optional)

We should now be green in Travis!


jbpoline:
thanks for removing these old datasets

PR #122, opened 2019-12-10
Update CONTRIBUTING.md with zenodo dataset upload procedure

joeyzhou98:


.github/CONTRIBUTING.md
```diff
@@ -40,7 +40,32 @@ for more information about PR evaluations.
 
 ## Adding data
 
-You can create a sub-dataset in the CONP repository as follows:
+You can create a sub-dataset in the CONP repository in two ways:
+
+1. Using Zenodo
+    1. Upload your dataset to https://zenodo.org with the specific keyword `canadian-open-neuroscience-platform` if the dataset is less than 50GB
```
dlq: Why less than 50GB?  (I know that Zenodo has a dataset file limit but do others?)

.github/CONTRIBUTING.md
```diff
@@ -40,7 +40,32 @@ for more information about PR evaluations.
...
+    2. If you set the dataset as restricted, create a personal token via Applications > Personal access tokens > New Token > Check all scopes > Create. Then send the token via email to tglatard@encs.concordia.ca
```
dlq: Why send the token to a random email address of someone I may not know?  (Again, of course, I know who Tristan is but this may seem weird to some random person who wants to add their data.)

.github/CONTRIBUTING.md
joeyzhou98: image
It seems like we can get more than 50GB but we would need to contact them

.github/CONTRIBUTING.md
joeyzhou98: The only way we can access the restricted dataset is if we get that token, do you have any recommendations to how to ask them to send the token more safely to us?

.github/CONTRIBUTING.md
joeyzhou98: I updated the documentation to reflect that option

.github/CONTRIBUTING.md
glatard: Added more info. We could use a support@conp.ca email address when we have one.

Issue #74, opened 2019-08-28
Merge folders "investigators" and "projects" into a new folder called "datasets"


joeyzhou98:
## Purpose
Merge folders "investigators" and "projects" into a new folder called "datasets" in order to have better structural organization

## Context
I am currently writing a Zenodo crawler to allow people to more easily upload datasets via a more automatized way where they can just upload their dataset to Zenodo with a "canadian-open-neuroscience-portal" tag/keyword and my script will retrieve those datasets and open pull requests to this repository. Having a single folder for datasets would simplify both organization and easy of access for everyone

## Possible Implementation (optional)
Simply move all the datasets in "investigators" and "projects" into a new one called "datasets"

jbpoline:
I am in favor of this - but with no hierarchy, if the number of datasets gets large (a good problem to have) browsing in the portal may become cumbersome. This would be mitigated by a search system on tags.

joeyzhou98:
@glatard yes, I will take care of this next

PR #128, opened 2019-12-13
Updating a dataset manual for CONTRIBUTING.md

gi114:
Addition of 'Updating a CONP dataset' workflow in CONTRIBUTING.md


emmetaobrien:
This document mentions the current list of modules in CONP as under CONP-datasets/projects, but we do also have three Khan lab datasets currently under CONP-datasets/investigators.  This organisation was raised as an issue (https://github.com/CONP-PCNO/conp-dataset/issues/74) some time ago but we should probably revisit that, so that we can determine whether to correct the documentation accordingly.

PR #132, opened 2020-01-09
Add copy of dataset_schema.json

xlecours:
## Description
Adds a copy of https://github.com/datatagsuite/schema/blob/master/dataset_schema.json for our validator to run against datasets. This will allow for us to specify additional mandatory fields and alter properties description to explain how those value are used by the portal. (this later point is subject to discussion)

Note: I am not sure what to put in the `id` property at line#2

### DIFF
--- a/schemas/conp_dataset_schema.json
+++ b/schemas/conp_dataset_schema.json
@@ -1,0 +1,208 @@
+{
+    "id": "https://portal.conp.ca/schemas/conp_dataset_schema.json",
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "DATS Dataset Schema for CONP portal",
+    "description": "This is an altered copy of https://w3id.org/dats/schema/dataset_schema.json",
+    "type": "object",
+    "properties": {
+        "@context": {
+            "description": "The JSON-LD context",
+            "anyOf": [
+                {
+                    "type": "string"
+                },
+                {
+                    "type": "object"
+                },
+                {
+                    "type": "array"
+                }
+            ]
+        },
+        "@id": {
+            "description": "The JSON-LD identifier",
+            "type": "string", "format": "uri"
+        },
+        "@type": {
+            "description": "The JSON-LD type",
+            "type": "string", "enum": [ "Dataset"]
+        },
+        "identifier": {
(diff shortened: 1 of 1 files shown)


emmetaobrien:
It occurs to me, as a potential alternative here,  that if we forked https://github.com/datatagsuite/schema/ into CONP-datasets, our validator could straightforwardly navigate all the other schemas for elements included in the datasets (which would be consistent with the behaviour of the perl validation script I have been using) and then editing the dataset_schema.json in that repository; right now I am not seeing any need for us to modify any of the other schemas, but that seems an option we could want in future.

glatard:
@emmetaobrien in that case we could add that repo as a submodule.

emmetaobrien:
Just adding the existing repo as a submodule would not give us the flexibility to edit any of the schemata independently, I was thinking.

PapillonMcGill:
Effectively, we need to clone datatagsuite here and then we could include that new repo inside conp-datasets.

glatard:
In that case I think you should fork that repo first, then add it here as a sub-module. Otherwise, you're going to loose the history on these files.

PapillonMcGill:
fork of datatagsuite/schema done in https://github.com/CONP-PCNO/schema last Friday.
Still have to add CONP specific required fields, then add as a submodule to datasets.

xlecours:
@PapillonMcGill I am not convince that we need a fork of the whole repo. What we need is an altered version of the dataset_schema.json (an potentially 1 or 2 other schemas) for the validator to run against the DATS.json files.

emmetaobrien:
@xlecours: If we want the validator to recursively check every element within the dataset_schema for correct formatting, we do need all or almost all of the DATS schema to be available, and I think it would be a better idea to have our own copies of all of them than to have local copies of the ones we change and then have to look elsewhere.

If all we look at is the dataset_schema, then we don't have a means to check formatting of dates, publications, biological entities and so on.

xlecours:
@emmetaobrien I see.
That way Travis would clone our repo before running the validator.

xlecours:
closing this. Modification of the schema(s) should go to https://github.com/CONP-PCNO/schema

PR #149, opened 2020-01-20
Adding CONP_status to DATS.json

emmetaobrien:
## Description

This PR adds one additional field to DATS.json for 10 of the datasets in CONP (1000GenomesProject, Brainspan, celltypes, mm_neo_atlas, multicenter-phantom, nhpatlas, preventad-open, refseq, SIMON-dataset and visual-working-memory), *extra_properties->CONP_status*, to indicate which category the dataset should be labelled as in the CONP portal.   Valid values are "CONP", "Canadian" or "external".  Equivalent values have been independently added to the six openpain datasets in separate PRs.

~~The values included are reasonable best estimates based on other metadata, and are not meant to be final.  The purpose of submitting this PR before receiving final confirmed values is to have the data structure in place so that front-end development can proceed; the expected outcome is that it should be possible to develop the desired visual distinction between the three categories in the interface even if category assignations of individual datasets needs correction later.~~ Final values have now been obtained and entered in the *extraProperties->CONP_status* field as appropriate

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -7,12 +7,13 @@ addAssignees: true
 # A list of reviewers to be added to pull requests (GitHub user name)
 reviewers: 
   - jbpoline
-  - paiva
   - edickie
   - glatard
   - dlq
   - PapillonMcGill
-  - shots47s
+  - emmetaobrien
+  - xlecours
+
 
 # A list of keywords to be skipped the process that add reviewers if pull requests include it 
 skipKeywords:
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 venv/
 tests/test*
 !tests/test_zenodo_crawler.py
+!tests/test_dats_validator.py
 __pycache__/
 .DS_Store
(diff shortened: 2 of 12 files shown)


PapillonMcGill:
Naser will provide the list this afternoon. We could wait until then to correct/approve.

emmetaobrien:
Sounds good to me.

PapillonMcGill:
How to specify that one of the extraProperties' array item must contain a "CONP_status"? 

If we had another fields outside of the array, we will break backward compatibility. But as done in this PR to preserve backward compatibility, I don't see how it is possible to make it required in the schema.

emmetaobrien:
I have now received Naser's assignments for each dataset and updated the repository accordingly.  While doing so I noticed that .gitmodules for some of the public datasets pointed to copies in my repository rather than at conpdatasets so I have corrected this also.

gi114:
Hi @PapillonMcGill which DATS validator are you currently using ? Would you share it with me so I can catch up on that. In case we might have to edit the validator to approve that item.

PapillonMcGill:
@gi114 The validator is https://github.com/CONP-PCNO/dats-validator

PR #156, opened 2020-01-23
Update in PERFORM dataset

glatard:
## Description

This PR replaces the PERFORM dataset created by @conp-bot by a clone hosted in my own account. It is a temporary solution until the Zenodo crawler is fixed. Please review in particular the DATS model in the new dataset and make sure it is complete enough.

## Checklist

Mandatory files and elements:

Functional checks:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -63,3 +63,6 @@
 [submodule "scripts/dats_validator/conp-dats"]
 	path = scripts/dats_validator/conp-dats
 	url = https://github.com/CONP-PCNO/schema
+[submodule "projects/PERFORM_one_control"]
+	path = projects/PERFORM_one_control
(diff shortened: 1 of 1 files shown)


emmetaobrien:
I had also prepared a clone in case the crawler did not work, but am as happy to go with yours.  The only minor query I have is whether using the accented "e" in Montreal could cause problems, I do not think we have tested that elsewhere, but I am not seeing anywhere I think that is likely.

xlecours:
@emmetaobrien I think it is fine. If not, I will add charset to https://github.com/CONP-PCNO/conp-portal/blob/master/app/search/models.py#L14

glatard:
Closing this as this is now a duplicate of https://github.com/CONP-PCNO/conp-dataset/pull/159

PR #161, opened 2020-01-24
fix license for preventad-open

PapillonMcGill:
## Description
fix dats.json for license for preventad-open


emmetaobrien:
The licenses filed looks right, but the name of the file is not capitalised.

PR #162, opened 2020-01-27
Integrate DATS validator in Travis tests

joeyzhou98:
## Description
- TravisCI tests now run the DATS validator on all DATS.json files in each submodule under directories "projects" and "investigators"
- Need to fix all DATS files before merging this PR so that the build stays passing


xlecours:
This seems to point out errors in our DATS files. Which is a good thing because there are not valid according to the schemas.

scripts/crawl_zenodo.sh
```diff
@@ -19,5 +19,5 @@ touch ${TOUCHFILE}
 
 (cd ${BASEDIR} && git pull main master)
 
-docker run --rm -u $UID:$UID -v $HOME:$HOME -e HOME=$HOME -v ${BASEDIR}:/workdir -w /workdir bigdatalabteam/git-annex python3 ./scripts/crawl_zenodo.py --force --verbose -z ${TOKEN_LIST} &>>${LOGFILE}
+docker run --rm -u $UID:$UID -v $HOME:$HOME -e HOME=$HOME -v ${BASEDIR}:/workdir -w /workdir bigdatalabteam/git-annex python3 ./scripts/crawl_zenodo.py --force --verbose -z ${TOKEN_LIST} &>>${LOGFILE} || \rm ${TOUCHFILE}
```
dlq: I'm unsure how adding `|| \rm ${TOUCHFILE}` does anything since the next line is `\rm ${TOUCHFILE}`

joeyzhou98: You are right, let me remove it

glatard:
As discussed today:
* @joeyzhou98 will fix the DATS model for the PERFORM dataset
* @emmetaobrien or @xlecours will fix the Prevent-AD and possibly OpenPain ones and PR to `joeyzhou98:integrate_dats_validator`

PR #163, opened 2020-01-27
Updating links to openpain repositories.

emmetaobrien:
This PR updates submodule links for openpain repositories to now point to forks of those under conpdatasets.  Two of these, placebo_1 and cbp_resting, have had DATS.json updated to correct failure to display correctly in the CONP portal.


emmetaobrien:
sorry, wrong branches.

PR #167, opened 2020-01-29
fix syntax error in DATS.json (Prevent-AD)

PapillonMcGill:
## Description
Fix schema syntax error for preventAD-open DATS.json file

## Checklist

Mandatory files and elements:
- [x ] A `README.md` file, at the root of the dataset
- [x ] A `DATS.json` file, at the root of the dataset
- [x ] If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset

Functional checks:
- [x ] Dataset can be installed using DataLad, recursively if it has sub-datasets
- [x ] Every data file has a URL
- [x ] Every data file can be retrieved or requires authentication
- [x ] `DATS.json` is a valid DATs model
- [x ] If dataset is derived data, raw data is a sub-dataset


glatard:
This should be closed and another PR should be made with the updates in https://github.com/conpdatasets/preventad-open/pull/12#event-3017404746 The sooner the better, as there are other tasks depending on this.

Issue #173, opened 2020-02-06
Crawler updates


glatard:
The following issues and feature requests were identified from the PR in https://github.com/CONP-PCNO/conp-dataset/pull/172 (dataset uploaded to Zenodo by @emdupre). 

# Bugs 

## DATS file
In:
```
"distributions": {
        "format": "txt",
        "size": 15461,
        "unit": {
            "value": "B"
        },
        "access": "https://zenodo.org/record/3647611"
    }
```
* [x] The data size is incorrect: 15461B instead of 13.9GB
* [x] The formats should include txt, nii.gz, tsv and any other extension found in the data files

# Feature requests

## Improved README
* [x] Add the dataset description to the dataset README
* [x] If the description is in HTML, convert it to Markdown before adding it to the README, using for instance https://pypi.org/project/html2markdown/

## Faster crawling

10.9GB of data had to be downloaded by the crawler to process this dataset:
* [ ] The crawler should be migrated to a server with a better network connection (offline discussion in progress)
* [x] The crawler should add URLs to the dataset without having to download them (`git-annex addurl --fast ...` ) except when these URLs point to zip archives that must be unpacked.

## Other
* [x] Add https://about.zenodo.org/static/img/logos/zenodo-gradient-round.svg as dataset logo
* [x] Add timestamps to the crawler logs:
```
[INFO] Downloading ...
```
should be:
```
[ <date-time> -- INFO ] Downloading ...
```

PR #176, opened 2020-02-07
Zenodo crawler fixes and improvements

joeyzhou98:
## Description
- Generated DATS.json now correctly displays total size and possible file formats
- Generated DATS.json contains zenodo logo as extra property
- Generated README.md contains dataset description in markdown format
- Crawling non-archive files (non zip files) doesn't require downloading (faster crawling)
- Time stamps in datalad logs are configured to be displayed when crawling
- Refactoring code so there is less redundant saving/committing/pushing

## Related issues (optional)
#173

### DIFF
--- a/scripts/crawl_zenodo.py
+++ b/scripts/crawl_zenodo.py
@@ -300,22 +300,21 @@ def create_new_dataset(dataset, token, force, username):
 
     for bucket in dataset["files"]:
         download_file(bucket, d, dataset_dir)
-    d.save()
 
     # Create DATS.json if it doesn't exist
     if not os.path.isfile(os.path.join(dataset_dir, "DATS.json")):
         create_new_dats(dataset_dir, os.path.join(dataset_dir, "DATS.json"), dataset)
-        commit_push_file(dataset_dir, "DATS.json", "[conp-bot] Create DATS.json", token)
+
+    # Create README.md if doesn't exist
+    if not os.path.isfile(os.path.join(dataset_dir, "README.md")):
+        create_readme(dataset, dataset_dir)
 
     # Add .conp-zenodo-crawler.json tracker file
     create_zenodo_tracker(os.path.join(dataset_dir, ".conp-zenodo-crawler.json"), dataset)
-    commit_push_file(dataset_dir, ".conp-zenodo-crawler.json", "[conp-bot] Create .conp-zenodo-crawler.json", token)
 
+    # Save all changes and push to github
+    d.save()
     d.publish(to="github")
-
-    # Create and push README.md if doesn't exist to github repo
-    if create_readme(dataset, dataset_dir):
-        commit_push_file(dataset_dir, "README.md", "[conp-bot] Create README.md", token)
 
     # Add description to Github repo
     add_description(token, repo_title, username, dataset)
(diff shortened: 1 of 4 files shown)


scripts/crawl_zenodo.py
```diff
@@ -432,8 +445,15 @@ def create_readme(dataset, path):
...
-Crawled from Zenodo"""
-                    .format(dataset["title"], dataset["doi_badge"]))
+Crawled from Zenodo
+
+## Description
+
+{2}""".format(
+                dataset["title"],
+                dataset["doi_badge"],
+                html2markdown.convert(dataset["description"]).replace("\n", "<br />"))
```
glatard: Not all descriptions will be HTML, is there a way to check that, or do you confirm that it's harmless to convert from HTML if the description is just plain text?

scripts/crawl_zenodo.py
```diff
@@ -241,17 +242,21 @@ def get_zenodo_dois(stored_tokens, passed_tokens, verbose=False):
...
-            "distributions": {
-                "format": files[0]["type"] if len(files) > 0 and "type" in files[0].keys() else None,
-                "size": files[0]["size"] if len(files) > 0 and "size" in files[0].keys() else None,
-                "unit": {"value": "B"},
-                "access": dataset["links"]["html"]
-            },
+            "keywords": metadata["keywords"] if "keywords" in metadata.keys() else [],
+            "distributions": [
+                {
+                    "formats": list(set(map(lambda x: x["type"], files))) if len(files) > 0 else None,
```
glatard: @emmetaobrien or @zxenia, do you confirm that `formats` can be a list in the DATS model?

scripts/crawl_zenodo.py
```diff
@@ -286,13 +291,16 @@ def create_new_dataset(dataset, token, force, username):
...
+    d.config.add("datalad.log.timestamp", "true")
+    d.save()
 
     r = d.create_sibling_github(repo_title,
                                 github_login=token,
                                 github_passwd=token)
 
     for bucket in dataset["files"]:
         download_file(bucket, d, dataset_dir)
+    d.save()
```
dlq: Why this `d.save` and the one on line 295?

scripts/crawl_zenodo.py
zxenia: @glatard it is a list, yes!

tests/test_zenodo_crawler.py
```diff
@@ -39,7 +39,8 @@ def mock_zenodo_query():
                             "links": {
                                 "self": "https://www.test.file.com/nofilehere"
                             },
-                            "type": "json"
+                            "type": "json",
+                            "size": 45346
```
dlq: Should this be a fixed number - even in a mocked test?  I'd feel better if this were computed somehow so that you don't have to remember to update this if the test changes.

scripts/crawl_zenodo.py
joeyzhou98: Yes I tested it and if the input to convert is regular string, it will just return the string as is

scripts/crawl_zenodo.py
joeyzhou98: You are right, let me refactor the code

tests/test_zenodo_crawler.py
joeyzhou98: I don't think we need to have this computed since the file size will always be available via Zenodo api

PR #178, opened 2020-02-07
fix khanlab submodules url

xlecours:
## Description
This is to fix [ERROR] Failed to clone from all attempted sources occuring for khanlab submodules


xlecours:
@glatard This should have been caught by the datalad install test right? I will look at Travis logs

glatard:
yes it should have, not sure why it didn't

Issue #170, opened 2020-02-05
Dataset download is not properly tested


glatard:
Some datasets, for instance "SIMON-dataset", pass tests while their data cannot be accessed. This seems to be due to (1) a bug in the recursion in the dataset dirs in `tests/functions.py`, (2) the fact that dataset download is only tested with probability `odds`, and there is no guarantee that exactly `num_files` files will be tested.

## Expected Behavior

Exactly `num_files` files should be tested for download, assuming that the dataset has at least `num_files` files. 

## Current Behavior

Datasets may pass the tests with no file being tested.

## Related issues

"SIMON-dataset" cannot be downloaded but it passes the tests. @xlecours @jbpoline FYI.

## Possible fix

Remove the random selection of files to be tested. Instead:
1. Get the list of all annexed files in the dataset
2. Shuffle this list
3. Test the first `num_files` files found in the list, or all the files in the list if the list has less than `num_files` elements
@joeyzhou98, this takes priority over https://github.com/CONP-PCNO/conp-dataset/pull/99

PR #180, opened 2020-02-08
TravisCI test more stringent (now tests download more thoroughly)

joeyzhou98:
## Description
Tests now use git annex to get a list of annexed files and randomly chooses a number of them to be tested instead of recursing into the directories and using `random() < odds` to calculate when to test files

## Related issues (optional)
#170

### DIFF
--- a/tests/create_tests.py
+++ b/tests/create_tests.py
@@ -8,7 +8,7 @@ template = Template("""from functions import examine
 
 
 def test_$clean_title():
-    assert examine('$path') == 'All good'
+    assert examine('$path')
 """)
 
 for dataset in submodules:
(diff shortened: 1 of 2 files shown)


joeyzhou98:
Need to merge #179 before TravicCI can run the test properly

joeyzhou98:
@glatard yes I am investigating this issue as behavior differs on my machine

glatard:
Update: tests currently pass when download timeouts (after 3 secs). The timeout should be increased to 30s, and tests should fail when timeout is reached, unless authentication message is present.

glatard:
Closing this, as this is now integrated in #197

PR #186, opened 2020-02-12
Expiring old branches in git-annex, and minor README update.

emmetaobrien:
## Checklist

Mandatory files and elements:
- [x] If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset

Functional checks:
- [x] If dataset is derived data, raw data is a sub-dataset


emmetaobrien:
This PR is primarily to update each repository using ```git annex expire```,  following advice from @yarikoptic, to prevent copies that no longer exist (eq, copies on defunct VMs) from appearing in ```git annex info```.  Any repository that has been inactive for more than thirty days has been expired in this update.  Any of these repositories that turn out to need restoring can be restored using the ```git annex semitrusted``` command.

emmetaobrien:
Also, the description template for this PR was not editable other than by ticking check boxes, hence the details in the above comment.

PR #197, opened 2020-02-18
Major changes of dataset tests suite

mathdugre:
## Description
Currently the tests have are not working properly. Those changes aims to allow datasets requiring authetication to be tested and to fix the procedure to verify the availability of the datasets remotes.

### Handling authentication
Currently, the tests cannot validate the proper functioning of datasets that require authentication.

The suggested changes allows TravisCI to authenticate to the dataset remote and proceed with the download verification. The Travis admin will have to add the appropriate credentials in TravisCI secrets; i.e. `${PROJECT }_USERNAME`, `${PROJECT }_PASSWORD`, and `${PROJECT }_LORIS_API`.
A limitation to the implementation is that TravisCI secret can only be used on the for code on the CONP-PCNO repository. That is, authenticated datasets will be assume to be working and will only be trulyy tested once they are merge in the CONP-PCNO repository.

### Download procedure
As of now, some datasets pass the tests suite although they do not function properly. This is due to the large download done to verify if the dataset remote is available; when the download timeout we currently assume the dataset works. Moreover, the number of files downloaded to test the dataset availability is not consistent due to the distribution choseen to download files randomly; this was some time resulting in no files to be downloaded and the dataset was assumed to be available.

To prevent dataset file downloads to timeout, a dataset files download in ascending order of of file size instead of randomly determining of a file should be downloaded. This changa also mkaes sure that a deterministic number of files is downloaded each times.


tests/functions.py
```diff
@@ -28,6 +30,25 @@ def raise_timeout(signum, frame):
     raise TimeoutError
 
 
+def generate_datalad_provider(loris_api):
+
+    # Regex for provider
+    re_loris_api = loris_api.replace(".", "\.")
+
+    with open("~/.config/datalad/providers/loris.cfg", "w") as fout:
```
glatard: Use `os.path.join` instead of `/` and `os.path.expanduser('~')` instead of `~`.

tests/functions.py
```diff
@@ -56,17 +77,38 @@ def recurse(directory, odds):
...
 
     return "All good"
 
 
-def examine(dataset):
+def examine(dataset, project):
+
+    # If authentication is required and credentials are provided then add credentials
+    # to the keyring and create a provider config file.
+    # Note: Assume a loris-token authntification.
```
glatard: authentication (typo)

tests/functions.py
```diff
@@ -56,17 +77,38 @@ def recurse(directory, odds):
...
-                msg = api.get(path=full_path, on_failure="ignore", return_type="item-or-list")
-                
+            with timeout(3):
+                msg = api.get(
+                    path=full_path, on_failure="ignore", return_type="item-or-list"
+                )
+
                 # Check for authentication
-                if isinstance(msg, dict) and msg["status"] == "error" and "unable to access" not in msg["message"].lower():
-                    return "Cannot download file and didn't hit authentication request for file: " + full_path
```
glatard: This needs to be updated as follows:
1. If dataset doesn't require authentication (check that from DATS model) then just print "Cannot download file (dataset doesn't require authentication)".
2. If dataset requires authentication then print "Cannot download file (dataset requires authentication, make sure that variables X, Y and Z are defined in Travis)"

PR #198, opened 2020-02-19
Updating DATS.json file format to include access->authorizations

emmetaobrien:
This PR updates the DATS.json files for 22 datasets to include the field  _distributions->access->authorizations_, which can contain the values "public", "registered" or "private". if this field is absent, the desired behaviour is for the value to be treated as "public".  Relevant files in conp-documentation are being updated appropriately.

This PR updates every dataset in CONP-PCNO/conp-dataset except for PERFORM and Learning_Naturalistic_Structure (obtained via conp-bot) and mcgill-emc-rna-seq-experiment (in @zxenia's repository).

PR #211, opened 2020-02-24
Updating prevent-ad

xlecours:
## Description
Updating PREVENT-AD to latest commit
This includes the renaming of image.json files


emmetaobrien:
@xlecours: The PreventAD DATS.json file in this commit is not compatible with the format update introduced in https://github.com/CONP-PCNO/conp-dataset/pull/198, it needs distributions->access->authorizations.  multicenter-phantom would be the best DATS.json to use as a template here.

PR #231, opened 2020-03-11
Updating 1000GenomesProject dataset

emmetaobrien:
This PR updates the 1000Genomes Project dataset. replacing links that were broken by a change of directory structure at the source ftp site, and adding 9 small supplementary files.


emmetaobrien:
Downloading from the EBI ftp site was unpredictably running extremely slow, so I have regenerated the links to point to a copy at NCBI which is behaving more reliably.

PR #99, opened 2019-10-14
Make Travis.CI only test submodules that differ (not ready, please do not merge)

joeyzhou98:
## Description
Current dataset testing procedure is unscalable, as it tests every single existing dataset present in repository, see related issue. Proposed solution is to test datasets/submodules that have changes from the incoming branch vs CONP-PCNO/conp-dataset:master in the PR:
1) Make a shallow clone of the master branch
2) Make a git diff with the incoming PR
3) From the diff, identify the datasets that are touched by the PR
4) For each touched dataset, run the tests

Detailed procedure is:
1) Shallow clone incoming PR branch
2) Shallow clone CONP-PCNO/conp-dataset:master in another directory
3) Install dependencies
4) `datalad install -r` both repositories
5) `git diff --name-only` both repositories and pipe file names output to `tests/diff.txt`
6) Clean `diff.txt` from numerous lines of `dev/null`
7) `tests/create_tests.py` now checks for existence of `diff.txt` and if it exists, it will create tests only for the submodules that have changes by reading and comparing from `diff.txt`
8) If `diff.txt` does not exist, for example when the user want to test their dataset locally, `tests/create_tests.py` creates tests for every single submodule so that the user can choose which ones to run
9) Run all tests

## Related issues (optional)
#89

PR #232, opened 2020-03-11
Minimal testing of datasets

mathdugre:
## Description
The tests suite currently validates every datasets on each pull request.
The objective of these changes is to perform the minimal amount of test required for futur pull request. Depending on the files a PR modifies it will either (1) run all datatset tests as before, (2) run only a sample of the dataset tests, or (3) none of the dataset tests.

Functional checks:
- [x] Added unit-test

## Related issues
PR #99

### DIFF
--- a/tests/create_tests.py
+++ b/tests/create_tests.py
@@ -1,8 +1,57 @@
+import os
 from string import Template
+from typing import List
 from git import Repo
+import requests
 
 
-submodules = list(map(lambda x: x.path, Repo(".").submodules))
+def minimal_tests(datasets: List[str], pr_files: List[str]):
+    """Return the minimal test set for a pull request.
+
+    Parameters
+    ----------
+    datasets : List[str]
+        List of datasets in the repository.
+    pr_files : List[str]
+        List of files modified by the pull request.
+    
+    Returns
+    -------
+    [type]
+        [description]
+    """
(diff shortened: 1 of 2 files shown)


tests/create_tests.py
```diff
@@ -1,19 +1,86 @@
...
+    ----------
+    datasets : List[str]
+        List of datasets in the repository.
+    pr_files : List[str]
+        List of files modified by the pull request.
+    
+    Returns
+    -------
+    [type]
+        [description]
```
glatard: Please explain in more details what the function returns.

tests/create_tests.py
```diff
@@ -1,19 +1,86 @@
...
+    ]
+
+    # No need to do tests when no modification are brought.
+    if len(pr_files) == 0:
+        return []
+
+    modified_datasets: List[str] = []
+    for pr_filename in pr_files:
+        for dataset in datasets:
+            if dataset in pr_filename:
```
glatard: Test should be more stringent.

glatard:
@dlq, you might want to have a look at this, as you had reviewed #99.

tests/create_tests.py
mathdugre: I clarified the function return and description in docstring.

PR #237, opened 2020-03-25
Circleci

mathdugre:
## Description
Currently, the continuous integration is performed on TravisCI, however, it is causing issues since the TravisCI is unreliable for FTP request. The current changes propose CircleCI as an alternative.

## Checklist
- [x] Configuration of CircleCI
- [x] Caching build dependencies
- [x] Running tests in parallel
- [x] Caching datasets

## Related issues (optional)
#197

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,17 +1,18 @@
 version: 2
 jobs:
   build:
-    working_directory: ~/conp-dataset
+    working_directory: /src/conp-dataset
     docker:
-      - image: mathdugre/conp-dataset
-    parallelism: 2
+      - image: mathdugre/conp-dataset:2.0
 
     steps:
-      - checkout
+      - checkout:
+          path: /src/conp-dataset
+
       - restore_cache:
           keys:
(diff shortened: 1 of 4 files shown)


xlecours:
I am not familiar with circleci. Can you briefly explain why it is a desirable alternative/addition ?

mathdugre:
Hi @xlecours, the main reason to consider CircleCI as an alternative to TravisCI was because it is more reliable for FTP connections. In #197 there were some datasets that were failing the test due to a download timeout while they were downloading a file of few Bytes. @glatard found that all those datasets were using an FTP connection. This issue is explained in more details [here](https://blog.travis-ci.com/2018-07-23-the-tale-of-ftp-at-travis-ci).

glatard:
I think we should merge this, configure Circle-CI, and then keep both Travis and Circle on for a few weeks, until we can check that Circle works as expected. Do you agree @xlecours?

xlecours:
ok :) Sorry for the response delay

PR #246, opened 2020-03-27
Implement OSF crawler

joeyzhou98:
## Description
- Add OSF platform crawler
- Does not support public/restricted datasets yet

### DIFF
--- a/scripts/crawl.sh
+++ b/scripts/crawl.sh
@@ -16,6 +16,6 @@ test -f ${TOUCHFILE} && (echo "Another crawling process is still running (${TOUC
 # We are in the protected section
 touch ${TOUCHFILE}
 
-cd ${BASEDIR}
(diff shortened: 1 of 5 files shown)


scripts/crawl.sh
```diff
@@ -16,6 +16,6 @@ test -f ${TOUCHFILE} && (echo "Another crawling process is still running (${TOUC
 # We are in the protected section
 touch ${TOUCHFILE}
 
-cd ${BASEDIR}
+cd ${BASEDIR} && git pull --no-edit main master
```
glatard: Have you considered doing that in the Python script instead?

joeyzhou98: Not sure if that is a good idea just in case `scripts/crawl.py` gets changed by the pull since if I do pull in `scripts/crawl.py` then any modification in `scripts/crawl.py` won't be seen/applied until the next run of the crawler

PR #256, opened 2020-04-14
test DATS.json reformat to specify PI

emmetaobrien:
This PR adds a sample modified DATS.json for the mm_neo_atlas dataset which assigns "Principal Investigator" to `roles->value`(edited for error) for the appropriate individual under `creators->name`, so that interface development can proceed while other datasets are updated.

Applying this change will require `creators->name` to be formatted as a list of individual name entries in all cases, where at the moment it is sometimes a single entry containing a list of authors.


cmadjar:
@glatard @edickie Do you want to take a look at the changes before I merge?

PR #273, opened 2020-04-19
Handle test of submodules in dataset

mathdugre:
## Description
The current test do not allow to test the submodules within a dataset; such as [visual-working-memory](https://github.com/conpdatasets/visual-working-memory/tree/7ad5fd7fd950494e5567dc13aa760a5167ba6795) dataset.
Allowing to test a dataset submodules improve the test coverage while preventing some dataset to be skipped during the tests.

### DIFF
--- a/tests/functions.py
+++ b/tests/functions.py
@@ -264,7 +264,7 @@ def examine(dataset, project):
         ]
 
     if len(filenames) == 0:
-        pytest.skip(f"WARNING: {dataset} No files found in the annex.")
+        return True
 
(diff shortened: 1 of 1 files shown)


glatard:
Do I get it well that files in submodules are added to the files to be (possibly) tested, but that no README or DATS model is required in sub-modules? (that's what we want, just checking)

mathdugre:
@glatard, Yes, It finds all submodules in a dataset then add their annexed files to the list of file to sample from.

PR #291, opened 2020-05-04
Template for broken dataset.

mathdugre:
## Description

The proposed change is to add an issue template to report problem when installing a dataset.


mathdugre:
@glatard do you think anything should be added?

PR #293, opened 2020-05-06
Crawler bug fixes and feature additions

joeyzhou98:
## Description
- When crawling datasets from Zenodo, when the crawler detects the presence of a contributor of the type "Project Leader", it will map it to the role "Principal Investigator" in the DATS.json file. It only allows the role of Principal Investigator for one creator for the entire list of creators
- OSF crawler now adds total file size, unit and number in generated DATS file
- OSF crawler now adds homepage link in DATS file

## Related issues (optional)
https://github.com/CONP-PCNO/conp-portal/issues/254


emmetaobrien:
That looks entirely reasonable to me.

glatard:
@joeyzhou98  could you test it on a sample dataset before we merge?

joeyzhou98:
@glatard it is tested and working well on the test dataset

Issue #294, opened 2020-05-08
Some subdatasets point to no longer existing commits


yarikoptic:
I am trying to establish automated updates of this superdataset on http://datasets.datalad.org/?dir=/conp-dataset .  Some sub-datasets  cannot be updated to the versions (commits) known to the super-dataset, e.g.:
```(git)smaug:/mnt/btrfs/datasets/datalad/crawl/conp-dataset[master]git
$> git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   projects/BigBrain_3DClassifiedVolumes (new commits)
        modified:   projects/BigBrain_3DSurfaces (new commits)
        modified:   projects/multicenter-phantom (new commits)

no changes added to commit (use "git add" and/or "git commit -a")

$> git submodule update
fatal: remote error: upload-pack: not our ref 95632a4cd4d889581aa46984fb80a3cb28603a89
fatal: the remote end hung up unexpectedly
Fetched in submodule path 'projects/BigBrain_3DClassifiedVolumes', but it did not contain 95632a4cd4d889581aa46984fb80a3cb28603a89. Direct fetching of that commit failed.
```
I guess because original subdatasets might have had new versions force-pushed thus killing some previous history.

glatard:
Hi @yarikoptic, thanks for pointing this out! Do you confirm that the changes in #295 and #296 solve the issue?

yarikoptic:
I think it is all good now, thank you!

PR #295, opened 2020-05-08
Update multicenter-phantom submodule

emmetaobrien:
This PR updates the multicenter-phantom submodule to address https://github.com/CONP-PCNO/conp-dataset/issues/294.


emmetaobrien:
Travis is failing on the similar problem with BigBrain derived datasets before it gets to multicenter-phantom, but https://github.com/CONP-PCNO/conp-dataset/pull/296 should fix that and allow full testing of this PR.

PR #317, opened 2020-05-27
Auto build and push Docker image for the repository

mathdugre:
## Description

Those changes allow building and pushing a Docker image of the `conp-dataset` every time something is pushed to the master branch.
This has two main purposes:
* Offering a base environment to install (or create) datasets.

```bash
docker run --rm -it mathdugre/conp-dataset:latest
```

*  Local environment for testing changes to datasets.

```bash
docker run --rm -itd \
    -v path/to/dataset/folder:/conp-dataset/projects/${DATASET_NAME}
    --name conp-dataset
    mathdugre/conp-dataset:latest

docker exec conp-dataset python tests/create_tests.py # Only required when a new dataset is added.
docker exec conp-dataset pytest tests/tests_projects_${DATASET_NAME}'
```

## Checklist

- [x] Merged #316
- [ ] Add `DOCKERHUB_USERNAME` secret to repository
- [ ] Add `DOCKERHUB_TOKEN` secret to repository

## Related issues (optional)

#9 Use case for this PR.


mathdugre:
@glatard, could you review this PR when you have some time. I think those changes could be useful for testing new dataset locally as well as having a containerized environment to download datasets.

glatard:
The PR looks excellent to me, thanks @mathdugre! Before we merge, we should configure the conp Docker organization in CircleCI, so that the container gets pushed there. 

@cmadjar, @emmetaobrien or @xlecours, there is already a conp organization on DockerHub (https://hub.docker.com/u/conp) but I don't have access: do you know how is the owner and if so, can I be co-owner so that I can use my Docker credentials in Circle?

cmadjar:
@glatard I was not aware that existed. @xlecours @emmetaobrien do you know?

emmetaobrien:
I have not worked with this; @dlq might have been involved with setting it up, I think.

glatard:
@paiva, are you the owner of https://hub.docker.com/u/conp and if so, could you add me (username: glatard) as owner?

glatard:
OK, we now have access to the conp organization on Docker, this should be working, let's merge and see!

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,24 @@
+on:
+  push:
+    paths:
+      - "projects/**"
+      - "tests/**"
```
mathdugre, line 5: Here, we might want to consider other paths to include for rebuilding the image; e.g. `**/requirements.txt` and `scripts/**`.
Another option would be to omit this part to rebuild on every push.
tag: @glatard

Issue #302, opened 2020-05-15
Khanlab/BigBrainMRICoreg file system check failure.


mathdugre:
Running `git annex fsck` on the dataset results in failures.

## Current Behavior

```bash
Cmd('git') failed due to: exit code(1)
  cmdline: git annex --fast --json --json-error-messages --quiet fsck
  stderr: '  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009asym/BigBrainHist-to-ICBM2009asym-nonlin.xfm
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009asym/BigBrainHist-to-ICBM2009asym-nonlin_grid_0.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009asym/BigBrainHist-to-ICBM2009asym-nonlin_grid_1.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009asym/BigBrainHist-to-ICBM2009asym-nonlin_grid_2.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009asym/BigBrainHist-to-ICBM2009asym-nonlin_grid_3.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009sym/BigBrainHist-to-ICBM2009sym-nonlin.xfm
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009sym/BigBrainHist-to-ICBM2009sym-nonlin_grid_0.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009sym/BigBrainHist-to-ICBM2009sym-nonlin_grid_1.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009sym/BigBrainHist-to-ICBM2009sym-nonlin_grid_2.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-ICBM2009sym/BigBrainHist-to-ICBM2009sym-nonlin_grid_3.mnc
  ** No known copies exist of HistTransformation/BigBrainHist-to-PD25/BigBrainHist-to-PD25-nonlin.xfm
  ** No known copies exist of HistTransformation/BigBrainHist-to-PD25/BigBrainHist-to-PD25-nonlin_grid_0.mnc
[...]

PR #318, opened 2020-05-27
Update conp-dataset for updates to khanlab-datasets/BigBrainMRICoreg

emmetaobrien:
This PR updates conp-dataset to point to the version of the BigBrainMRICoreg dataset that was updated recently in order to address https://github.com/CONP-PCNO/conp-dataset/issues/302.


.gitmodules
```diff
@@ -105,3 +102,6 @@
 [submodule "projects/Intracellular_Recordings_of_Murine_Neocortical_Neurons"]
 	path = projects/Intracellular_Recordings_of_Murine_Neocortical_Neurons
 	url = https://github.com/conp-bot/conp-dataset-Intracellular-Recordings-of-Murine-Neocortical-Neurons.git
+[submodule "projects/Khanlab/BigBrainMRICoreg"]
```
cmadjar, line 105: shouldn't this point to the conpdatasets organization instead of your username?

emmetaobrien: Fair point, that was an error in reinstalling the reinitialised dataset.  Will fix.

emmetaobrien: Have now updated to fix this.

PR #343, opened 2020-06-18
[WIP] Added cneuromod umbrella dataset

glatard:
## Description

This is the umbrella dataset of the Courtois Neuromod project, available [here](https://github.com/courtois-neuromod/cneuromod). The sub-datasets `anat`, `movie10`, `hcptrt` and `friends` are not public yet, but it should be quite easy to add them when they are. Dataset was already a DataLad repository, only a DATS file was required.

## Checklist

Mandatory files and elements:

Functional checks:
- [ x ] Dataset can be installed using DataLad, recursively if it has sub-datasets
- [ x ] Every data file has a URL
- [ x ] Every data file can be retrieved or requires authentication
- [ x ] `DATS.json` is a valid DATs model

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -112,3 +112,6 @@
 	path = projects/preventad-open-bids
 	url = https://github.com/conpdatasets/preventad-open-bids.git
 
+[submodule "projects/cneuromod"]
+	path = projects/cneuromod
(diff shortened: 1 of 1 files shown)


glatard:
It doesn't pass the tests, let's not merge this for now.

Issue #340, opened 2020-06-18
preventad-open-bids fails to download


glatard:
* Test build: 
https://app.circleci.com/pipelines/github/CONP-PCNO/conp-dataset/362/workflows/dc192141-f465-44c5-ae2d-179bd34bbb5b/jobs/703/parallel-runs/0?filterBy=FAILED

* Standard error:
```
________ TestDataset.test_has_valid_dats[projects/preventad-open-bids] _________
 
Dataset projects/preventad-open-bids doesn't contain a valid DATS.json.
 
------------------------------ Captured log call -------------------------------
 
ERROR    scripts.dats_validator.validator:validator.py:48 1 Validation error in extraProperties.6.values.0: Additional properties are not allowed ('parent_dataset_id' was unexpected)
 
_
```

* Help requested from:
@cmadjar @emmetaobrien 
Related issue: #338

cmadjar:
looks like the DATS error is not present anymore in circle CI. There is still the dataset timed out issue though.

@glatard should we close this one and open another one for preventad-open-bids with the timeout issue?

glatard:
I think we could just leave this one open and track the download error here.

glatard:
The download error is coming from the download test of files in the parent dataset (submodule), in spite of the parent dataset passing the tests. This is most likely because the parent dataset needs different credentials than the dataset, and CircleCI uses the same set of credentials for both.

I see two possible solutions:
1. We don't test submodules at all: if the submodule is in CONP, great, it's tested somewhere else; otherwise, too bad, but this is also not a CONP dataset so it's not in our scope.
2. We extend the tests to support different credentials for the dataset and subdatasets.

I'm leaning slightly toward solution 1, as it is way simpler than 2, and 2 is also a rabbit hole. Thoughts @emmetaobrien and @cmadjar?

cmadjar:
@glatard in theory, the credentials are identical for those two datasets since they are connecting to the same LORIS instance.

However, I agree with you and lean towards solution 1 too. Seems redundant to test a submodule that we already test elsewhere and there is nothing we can do if a submodule is not a CONP dataset.

emmetaobrien:
Option 1 seems reasonable to me too.

glatard:
great, I'll change that soon. @mathdugre had told me that testing submodules was going to create issues, I should have listened to him :)

cmadjar:
@glatard did you update the LORIS credentials for the tests of this dataset in the end?

[...]

PR #348, opened 2020-06-23
Remove testing of sub-datasets

glatard:
## Description

This removes the testing of sub-datasets, following the discussion in #340

## Related issue

#340


glatard:
this doesn't pass the tests, for instance:
```
projects/mcgill-emc-rna-seq-experiment/projects/mcgill-emc-rna-seq-experiment/RNA-Seq-signal-forward-bigWig
 path does not exist
 ```

PR #355, opened 2020-07-09
Updating "distribution" fields in "hasPart" entries.

emmetaobrien:
This PR is to correct` hasPart` entries in the 1000Genomes, Brainspan and celltypes datasets by adding the required values under distribution.

This PR is not yet ready for merging, as I am currently testing that the corrections to 1000Genomes pass CircleCI correctly before applying them to the other two datasets.


emmetaobrien:
OK, Brainspan and celltypes have now been updated so this is ready to go.

PR #372, opened 2020-07-24
Crawler result (An End-to-end System for Automatic Characterization of Iba1 Immunopositive Microglia in Whole Slide Imaging: Data)

gi114:
## Description
Created An End-to-end System for Automatic Characterization of Iba1 Immunopositive Microglia in Whole Slide Imaging: Data

## Checklist

Mandatory files and elements:

Functional checks:


scripts/crawl.py
```diff
@@ -64,15 +64,20 @@ def parse_args():
...
+        # OSFCrawler = OSFCrawler(github_token, config_path, verbose, force)
+        # OSFCrawler.run()
 
         if verbose:
-            print(os.linesep + "==================== OSF Crawler Running ====================" + os.linesep)
-        OSFCrawler = OSFCrawler(github_token, config_path, verbose, force)
-        OSFCrawler.run()
+            print(os.linesep + "==================== FRDR Crawler Running ====================" + os.linesep)
+        FRDRCrawler = FrdrCrawler(github_token, config_path, verbose, force)
+        FRDRCrawler.run()
```
glatard, line 80: The three crawlers should be kept active, maybe the scripts could be updated in a separate PR.

PR #378, opened 2020-07-29
Fix circleci errors

joeyzhou98:


.circleci/config.yml
```diff
@@ -59,9 +59,10 @@ jobs:
           name: Run Tests
           command: |
             . ~/venv/bin/activate
+            pip install gitpython==3.1.7
```
glatard, line 62: could you add that to requirements.txt instead?

joeyzhou98: I've tried that and it didn't work: https://app.circleci.com/pipelines/github/joeyzhou98/conp-dataset/12/workflows/a2c334f2-445b-4157-8e57-b6f515130341/jobs/33

glatard: ok then...

emmetaobrien:
@joeyzhou98 @glatard : Am I right in reading this as a potential fix for the broken tests that can't be merged because the tests are broken, and if so, do we need to temporarily relax the "required statuses must pass before merging" constraint in order to proceed?

joeyzhou98:
@emmetaobrien no I haven't figured out the issue yet but it seems like running the CircleCI build and test works fine when it is ran from my own CircleCI account but as soon as I switch to CONP-PCNO's account, it gives the same error as the other tests. So it is something to do with different organization settings in CircleCI and I am still investigating

glatard:
The `ModuleNotFoundError: No module named 'tests.template' seems to be due to the lack of `__init__.py` file in `tests`. I'm not sure how it was working before. Will PR the fix separately.

glatard:
Hi @emmetaobrien not sure what's going on yet, #379 fixes some errors but there is still one remaining. I think we should take the time to fix the tests instead of merging things that might be broken. Will look at it further tomorrow.

emmetaobrien:
I am absolutely in favour of not merging things that might be broken, I was just worried that it looked like we might have backed into a circular dependency here, and am glad to hear that is not the case.

glatard:
This should now be fixed by #379

PR #387, opened 2020-08-12
OSF Crawler now supports private datasets

joeyzhou98:
## Description
- OSF Crawler can now handle private datasets and files, see https://github.com/CONP-PCNO/conp-dataset/pull/386
- Pip package versions should be constant to avoid discrepancies


cmadjar:
@joeyzhou98 there are conflicts with master

joeyzhou98:
There is already an OSF unit test, but it doesn't test retrieving private datasets as they are a bit hard to mock and test

PR #401, opened 2020-08-25
Conp bot/an end to end system for automatic characterization of iba1 immunopositive microglia in whole slide imaging  data

gi114:
## Description
Created a _new_ PR for An End-to-end System for Automatic Characterization of Iba1 Immunopositive Microglia in Whole Slide Imaging Data

Mandatory files and elements:

Functional checks:


cmadjar:
@glatard @gi114 @emmetaobrien Can we make this PR work? I know @gi114 is not working with us anymore but it would be really sad to lose that work. How can we ensure this is done?

PR #428, opened 2020-09-18
Fixes the origin_country field of the DATS file

cmadjar:
## Description

This fixes the DATS.json file where the origin_coutry field was labelled origin_institution instead.


emmetaobrien (approved):
Thanks for catching my error.

Issue #432, opened 2020-09-23
Polygenic Risk Score dataset(s)


emmetaobrien:
An initial dataset of output data and summary documents was posted to Zenodo at https://zenodo.org/record/4042021 yesterday, with manually generated DATS.json; this did not get picked up by the Zenodo crawler and we are currently looking into why.

cmadjar:
The crawler currently crashes on this dataset due to the fact that the crawler expects a `conceptdoi` field associated to the dataset, however, a `conceptdoi` is attributed to a dataset only if there are no existing `doi` specified when uploading the data. Basically,
- if a user does not specify a DOI at the time of upload, Zenodo will create a Concept DOI and another DOI for the version of the dataset. The Concept DOI represents the dataset and all its versions.
- if a user specifies a DOI at the time of the upload, Zenodo will not create DOIs (neither concept DOI or version DOI).  

@emmetaobrien **The fact that the DOI is not a Zenodo DOI means that the files in those datasets cannot be edited** So if the DATS.json or README files need to be updated for those datasets, it cannot be done, or at least easily...
See first general question from https://help.zenodo.org/.

I will send a PR to modify the crawler code to take the DOI instead of the concept DOI when no concept DOI is available for the crawled dataset.

cmadjar:
The PR above should fix the crawler. Once this is merged, I will pull the code on the crawler server and that dataset should be caught by the crawlers then.

cmadjar:
https://github.com/CONP-PCNO/conp-dataset/pull/435 has been merged and the dataset now appears on the portal

PR #434, opened 2020-09-23
Fix Zenodo crawler for datasets with no concept DOI

cmadjar:
## Description

This fixes a bug encountered by crawling the dataset mentioned in the following issue: https://github.com/CONP-PCNO/conp-dataset/issues/432.

In Zenodo, a concept DOI is attributed to a dataset upon the creation of a dataset with no pre-existent DOI associated with it. Here is how the Zenodo DOI works:
- if a user does not specify a DOI at the time of a dataset upload on the Zenodo platform, Zenodo will create a "Concept DOI" and a "version DOI". The Concept DOI represents the dataset and all its associated versions.
- if a user specifies a DOI at the time of the upload, Zenodo will not create new DOIs (neither concept DOI or version DOI) and it will use the specified DOI the user gave instead.

In the case of the mentioned dataset above, the DOI was a Frontiers in Neuroscience DOI, therefore, there does not exist a "Concept DOI" for that dataset which is being used in the `DATS.json` file. 

This PR fixes the issue by either using the "DOI" field of the Zenodo API response for the dataset as identifier in the `DATS.json` file if the "Concept DOI" does not exist for the dataset.

## Related issues 

https://github.com/CONP-PCNO/conp-dataset/issues/432

### DIFF
--- a/scripts/Crawlers/ZenodoCrawler.py
+++ b/scripts/Crawlers/ZenodoCrawler.py
@@ -172,17 +172,19 @@ class ZenodoCrawler(BaseCrawler):
                                     "roles": [{"value": "Principal Investigator"}]}
                             )
 
+            identifier = dataset["conceptdoi"] if "conceptdoi" in dataset.keys() else dataset["doi"]
+
             zenodo_dois.append(
                 {
                     "identifier": {
-                        "identifier": "https://doi.org/{}".format(dataset["conceptdoi"]),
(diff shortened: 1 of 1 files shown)


cmadjar:
@emmetaobrien this should resolve the issue with the Zenodo crawler experienced with the polygenic risk score dataset. Do you want to take a look?

emmetaobrien (approved):
Makes sense to me in principle and execution, let's give it a go.

PR #439, opened 2020-09-29
Set Braincode datasets authorizations to private

zxenia:
## Description
Previously, all Braincode datasets had `'authorizations'` value set to `'Public'`, after discussion we decided that the correct value should be `'private'`. This PR sets all authorizations values to `'private'`.

The Mouse Image dataset was missing origin info, it is fixed with this PR too.


cmadjar (approved):
LGTM 👍

PR #442, opened 2020-10-01
Make the conp-dataset PATH configurable for the crawlers

cmadjar:
## Description

This modifies the code of the crawler so that the conp-dataset path can be configured and not hardcoded.

In `.conp_crawler_config.json`, the following needs to be added to the config file:
```
  "conp-dataset_path": "PATH TO THE conp-dataset DIRECTORY",
```

Then the scripts will load the ZenodoCrawler and OSFCrawler libraries based on that path.

NOTE: this config has already been added to the `.conp_crawler_config.json` file of `conp-bot@zenodo-crawl.acelab.ca`

### DIFF
--- a/scripts/crawl.py
+++ b/scripts/crawl.py
@@ -4,10 +4,6 @@ import argparse
 import os
 import sys
 import traceback
-sys.path.append(os.path.abspath(os.path.join(os.path.expanduser("~"), "conp-dataset")))
-sys.path.append(os.path.join("/data", "crawler", "conp-dataset"))
-from scripts.Crawlers.ZenodoCrawler import ZenodoCrawler
-from scripts.Crawlers.OSFCrawler import OSFCrawler
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


scripts/crawl.py
```diff
@@ -44,6 +40,20 @@ def parse_args():
...
+        raise Exception(
+            '"conp-dataset_path" not configured in ' + config_path + ','
+            'please configure it as follows: \n'
+            '  "conp-dataset_path": "PATH TO conp-dataset DIRECTORY",'
+        )
+
+    # import the crawler packages
+    sys.path.append(config['conp-dataset_path'])
+    global ZenodoCrawler
+    global OSFCrawler
```
glatard, line 53: Do we really need to add global variables? It's not great practice in general.

glatard (commented):
See comment on global variables. Otherwise, LGTM!

scripts/crawl.py
cmadjar: @glatard I was not too keen either about the global variables. I slept on it and found another way to import the classes directly in the __main__ function. Let me know what you think.

glatard (approved):

PR #445, opened 2020-10-02
Support updates for datasets that do not have concept DOIS

cmadjar:
## Description

Uploaded datasets to Zenodo with an existing DOI do not have a concept DOI and their DOI remains static even though they update a file (like for example DATS.json). In order to support the updated file for such datasets, we need to rely on the `updated` field of the Zenodo JSON response and compare it with the version that conp-dataset has.

The changes required:
- changes to the `.conp-zenodo-crawler.json` file to add:
```
    "zenodo": {
        "concept_doi": "4042020", 
        "version": "4042021", 
        "updated": "2020-10-01T12:42:44.535301+00:00"       <=======
    }
```
- creation of a new function `determine_if_update_is_necessary` that returns `True` or `False` and prints to the user what the script will do for the dataset (if verbose turned on). 
- addition of the following to the `DATS.json`:
```
                    "dates": [
                        {
                            "date": dataset["created"],
                            "type": {
                                "value": "Date Created"
                            },
                        },
                        {
                            "date": dataset["updated"],
                            "type": {
                                "value": "Last Updated"
                            }
                        }
                    ],
```

## Related issues (optional)

This is currently blocking the update of the ePRS dataset: https://github.com/CONP-PCNO/conp-dataset/issues/432


scripts/Crawlers/ZenodoCrawler.py
```diff
@@ -186,6 +192,20 @@ def get_all_dataset_description(self):
                         "identifier": "https://doi.org/{}".format(identifier),
                         "identifierSource": "DOI",
                     },
+                    "dates": [
```
cmadjar, line 195: @emmetaobrien does the dates part sound OK to you for the DATS.json?

cmadjar:
@glatard could you review the code? I tested the new function and the modification of the function `.conp-zenodo-crawler.json` in a separate script and it appears to behave as expected but I would prefer if you also reviewed the code :).

scripts/Crawlers/ZenodoCrawler.py
emmetaobrien: It looks entirely reasonable to me, and should be straightforward to generate automatically from the GUI too I would expect.

cmadjar:
Given that in the example dataset that failed being updated by the Crawler used a paper DOI instead of creating a new DOI for the dataset itself, this change is not required so closing the PR.

PR #446, opened 2020-10-07
Modifies the crawlers to download DATS.json and README.md files instead of annexing them if they are present in the uploaded dataset

cmadjar:
## Description

Some datasets will provide their own `DATS.json` and `README.md` files in Zenodo or OSF. This PR modifies the crawler so that instead of indexing those two files, it will instead download them so later on they can be read by the validator.

Note: I also removed the patch I issued last week to rename `DATS` to `DATS.json` since Zenodo now supports `.json` file uploads.


glatard (approved):
Looks very good to me!

PR #449, opened 2020-10-08
Recreate README.md if it does not exist

cmadjar:
## Description

In the case the dataset is updated, the README.md file is not recreated. I missed that part yesterday when I refactored the crawler so that it could use the DATS.json and README.md uploaded by users through Zenodo or OSF.

This fixes it and add the creation of README.md file if a dataset is updated (note: this is done by default for new datasets).

## Related issues (optional)

This should fix the current issue with PR #448


cmadjar:
@glatard if you could review, that would be great. Let me know if you think of someone else I could bug to review the crawler PRs. This one should be pretty straightforward to review though. Thanks!

emmetaobrien (approved):
Looking good.

PR #458, opened 2020-10-19
[OSF crawler] Get the size of annexed files from git-annex info if it is not present in the OSF API response

cmadjar:
## Description

OSF supports multiple data providers. In the case of Github data provider (and probably all non-OSF Storage data provider), the file size information is not available in the API response. This causes the crawler to crash with the following error:

```
Traceback (most recent call last):
  File "./scripts/crawl.py", line 84, in <module>
    OSFCrawlerObj.run()
  File "/data/crawler/conp-dataset/scripts/Crawlers/BaseCrawler.py", line 254, in run
    self.add_new_dataset(dataset_description, dataset_dir)
  File "/data/crawler/conp-dataset/scripts/Crawlers/OSFCrawler.py", line 207, in add_new_dataset
    dataset_size, dataset_unit = humanize.naturalsize(sum(dataset_size)).split(" ")
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

The proposed fix will compute the file_size using the `git annex info --bytes <file>` command and will append it to the `dataset_size` array.

## Related issue

This is currently blocking the addition of the CFMM-7T: MP2RAGE T1 mapping dataset.

### DIFF
--- a/scripts/Crawlers/OSFCrawler.py
+++ b/scripts/Crawlers/OSFCrawler.py
@@ -108,7 +108,10 @@ class OSFCrawler(BaseCrawler):
                 # append the size of the downloaded file to the sizes array
                 file_size = file['attributes']['size']
                 if not file_size:
-                    file_size = annex('info', '--bytes', os.path.join(inner_path, file["attributes"]["name"]))
+                    # if the file size cannot be found in the OSF API response, then get it from git annex info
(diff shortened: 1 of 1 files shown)


cmadjar:
@glatard @emmetaobrien ready for review when you get a chance.

Note: I tested the changes locally on the CFMM new dataset and it resolved the issue.

emmetaobrien (approved):
looks reasonable to me.

PR #461, opened 2020-10-19
[Crawler] Resolves the deprecation notice regarding Github token in URL

cmadjar:
## Description

Github is deprecating the API authentication through query parameters. Instead, they recommend users move the authentication in the header of the request.

This PR modifies our crawler to use authentication in the header of the request.

## Related issues (optional)

See Github documentation: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/


cmadjar:
@glatard ready for review :). Should fix the monthly notification we get from Github about the deprecation of API authentication through query parameters that we received today.

glatard (approved):
Looks perfect!

Issue #436, opened 2020-09-29
[validator] check that a DataLad subdataset exists when DerivedFrom is filled in DATS.json


cmadjar:
## Purpose

A derived dataset should always have the dataset it is derived from as a DataLad subdataset. Below is the example of how derived datasets should be organized (example of preventad-open-bids).

- DATS.json content for DerivedFrom example
```
    {
      "category": "derivedFrom",
	  "values": [
        {
          "value": "https://github.com/conpdatasets/preventad-open/tree/17dd9742a89e60f720f8276d79d917c54dd20695"
        }
      ]
    },
```

- Directory structure of a derived dataset (example of preventad-open-bids):
```
.datalad/
BIDS_dataset/
preventad-open @ 17dd974/
.gitattributes
.gitmodules
DATS.json
README.md
logo.png
```

=> add a check that verifies that the DataLad subdataset is indeed present and on the correct commit (based on the value in the DATS.json file)

zxenia:
@cmadjar 
In the example, `derivedFrom` points to a specific version of preventad-open dataset: `https://github.com/conpdatasets/preventad-open/tree/17dd9742a89e60f720f8276d79d917c54dd20695` (from Aug 11)
But currently, there is a newer version of preventad-open dataset (From Oct 13). Is this a desired behavior, to have a specific version in the tree here?
Does this mean that this dataset https://github.com/conpdatasets/preventad-open-bids is not derivedFrom from the latest version https://github.com/conpdatasets/preventad-open ?

cmadjar:
@zxenia I think the only file that changed between the two commits is the DATS file to add the Zenodo DOI. All the data files are identical. 

I guess I need to send a PR to update the preventad-open dataset in the preventad-open-bids dataset. Unless there is a better way to do that?

zxenia:
Then would it be a solution to point to a dataset repo in `derivedFrom` but not to its version in a tree?
for example:

```
{
      "category": "derivedFrom",
	  "values": [
        {
          "value": "https://github.com/conpdatasets/preventad-open"
        }
      ]
 }
```

cmadjar:
that seems reasonable to me. @emmetaobrien do you agree? If so, I will send a PR to fix that shortly :).

emmetaobrien:
This entirely makes sense to me.

PR #468, opened 2020-11-18
Add derivedFrom validation

zxenia:
## Description
The PR addresses the extraProperties category `derivedFrom` validation.

- The code validates `derivedFrom`  value which is expected to be a valid url: sends request and checks if the response status code is 200.

- The code is not bounded to the GitHub API because I assume we might have cases when the `derivedFrom` points to the dataset persistent identifiers on other data portals or doi.org.

- To avoid redundant calls response statuses are stored in in-memory cache dict.

## Related issues (optional)
#436

### DIFF
--- a/scripts/dats_validator/examples/invalid_dats.json
+++ b/scripts/dats_validator/examples/invalid_dats.json
@@ -2874,6 +2874,14 @@
           "value": "logo.png"
         }
       ]
+    },
+    {
+      "category": "derivedFrom",
+      "values": [
+        {
+          "value": "https://github.com/sdajdhajhdahdj"
+        }
+      ]
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


cmadjar:
@zxenia sorry, I am losing track. What is the status of that PR?

zxenia:
@cmadjar the ci tests are failing, but it's related to test_download

cmadjar:
I see, the two datasets that are failing the test_download are:
- mm_neo_atlas (like all other PRs since their server appears offline, see https://github.com/CONP-PCNO/conp-dataset/issues/481 currently under investigation)
- refseq_Homo_sapiens (like all other PRs but the following PR will fix it https://github.com/CONP-PCNO/conp-dataset/pull/479)

I am tempting on merging the PR since the two errors reported by Circle CI are known and expected. @emmetaobrien what do you think?

emmetaobrien:
Those two issues should both be fixed when https://github.com/CONP-PCNO/conp-dataset/pull/479 is merged, so let's do that first and then confirm this passes.

cmadjar (approved):

PR #469, opened 2020-11-27
Adding preventad-registered to the conp-dataset super dataset

cmadjar:
## Description

This will add the preventad-registered dataset to the conp-dataset. 

NOTE: a DOI will be created on Monday upon the official release of PREVENT-AD registered and this PR will be updated with the correct DOI once we have it.

## Checklist

Mandatory files and elements:

Functional checks:


cmadjar:
@emmetaobrien I sent the PR to see if the DATS file is valid. Monday will be the official release of the dataset and upon its release I will have a DOI which I will add to the DATS file on Monday.

emmetaobrien (approved):
Seems fine to me.

PR #484, opened 2020-12-14
[Zenodo crawler] Fix dictionary where the created and updated keys are stored in the Zenodo API response

cmadjar:
## Description

This fixes the call to the dictionary that stores the created and updated keys in the Zenodo API response.


emmetaobrien (approved):
Looks good to me.

Issue #419, opened 2020-09-11
Handle OSF component datasets and treat them as derived datasets


cmadjar:
## Purpose

Currently, the crawler treats OSF components as any other datasets. Instead it should link them to the original dataset.

Example: https://osf.io/xhvsk/ is a dataset with 2 components (https://osf.io/pny28/ and https://osf.io/px49j/) but no files outside the components

https://osf.io/xhvsk/ resulted in the PR https://github.com/CONP-PCNO/conp-dataset/pull/397 that got closed immediately as there were no files associated with the dataset sent in the PR. 

However, the two components associated to that dataset resulted in two PRs: https://github.com/CONP-PCNO/conp-dataset/pull/405 and https://github.com/CONP-PCNO/conp-dataset/pull/406 but they have no relation to the original dataset.

Need to modify the crawler so the created DATS files and relation between datasets can be taken into account.

Note to self: look at the [YODA principles](http://handbook.datalad.org/en/latest/basics/101-127-yoda.html) definition before handling parent/child definition as we may be using it wrong.

cmadjar:
After discussions at the CONP dev meeting of December 16th, 2020, the components will not be handled as derived dataset. Instead, the OSF crawler will be modified to grep the files from the OSF components (in addition to the files already present in the dataset) when components are available.

To avoid sending PRs for each components, the OSF crawler will also ignore datasets returned in the OSF API response that have a parent key since the content of the dataset is part of another dataset.

PR #486, opened 2020-12-17
Handling OSF components

cmadjar:
## Description

This modifies the OSF crawler to support components. 

- When a dataset contains OSF components (example: https://osf.io/xhvsk/), then the content of the components will be crawled along with the other files from the dataset.

- For datasets that have a parent key in the API response (a.k.a. the dataset is a component of another OSF dataset), the crawler will skip that dataset since its content is going to be part of the parent dataset linked.

Notes:
- #477 conp-bot PR is the result of the crawler modifications present in the PR
- #405 and #406 were two earlier PRs for components dataset. Since the files are now in #477, those two PRs have been closed.

## Related issues 

https://github.com/CONP-PCNO/conp-dataset/issues/419

### DIFF
--- a/scripts/Crawlers/OSFCrawler.py
+++ b/scripts/Crawlers/OSFCrawler.py
@@ -115,17 +115,25 @@ class OSFCrawler(BaseCrawler):
                 sizes.append(file_size)
 
     def _download_components(self, components_list, current_dir, d, annex, dataset_size):
+        # Loop through each available components and download their files
         for component in components_list:
             component_title = self._clean_dataset_title(component['attributes']['title'])
-            component_path  = os.path.join(current_dir, 'components', component_title)
-            os.makedirs(component_path)
+            component_inner_path = os.path.join('components', component_title)
+            os.makedirs(os.path.join(current_dir, component_inner_path))
             self._download_files(
                 component['relationships']['files']['links']['related']['href'],
-                component_path,
+                os.path.join(current_dir, component_inner_path),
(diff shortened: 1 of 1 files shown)


scripts/Crawlers/OSFCrawler.py
```diff
@@ -114,6 +114,27 @@ def _download_files(self, link, current_dir, inner_path, d, annex, sizes):
...
+            component_inner_path = os.path.join('components', component_title)
+            os.makedirs(os.path.join(current_dir, component_inner_path))
+            self._download_files(
+                component['relationships']['files']['links']['related']['href'],
+                os.path.join(current_dir, component_inner_path),
+                component_inner_path,
+                d,
+                annex,
+                dataset_size
+            )
```
glatard, line 130: what happens if the component has a sub-component? not sure if it's even possible, but if it is we should support it or at least print a warning that it is not supported.

glatard (commented):
Looks great! I just have one comment, see inline.

scripts/Crawlers/OSFCrawler.py
cmadjar: good point. Did not think of that possibility. I'll check if this is possible with OSF but I don't see why this would not be possible. Thanks!

cmadjar: confirmed. We can create subcomponents. Will work on modifying the code this PM to accommodate for that. :)

PR #488, opened 2020-12-23
change calgary-campinas link to point to conpdatasets

emmetaobrien:
This PR updates the calgary-campinas subdataset link, which is failing to update in the portal, to point to a fork in conpdatasets rather than emmetaobrien in case this makes a difference.


liamocn (approved):
LGTM!

Issue #499, opened 2021-01-15
Harmonization of the DATS.json date type field


cmadjar:
## Purpose

Harmonization of the date types values of the DATS.json files describing datasets.

Convention: date type values should be all lower case except for "CONP DATS fileset creation date" where the acronyms should be capitalized.

**Initial list of date types based on what is already present in the datasets:**
- CONP DATS JSON fileset creation date
- start date
- end date
- release date
- last update date

## Context

This is one subtask related to #454

## Possible Implementation (optional)

- [x] convert date types in the existing DATS files to comply with convention above for non-crawled datasets:
  - [x] mcgill-emc-rna-seq-experiment https://github.com/CONP-PCNO/conp-dataset/pull/497
  - [x] all 3 Khanlab datasets https://github.com/CONP-PCNO/conp-dataset/pull/492
  - [x] all 3 PREVENT-AD datasets #500
  - [x] all 3 braincode datasets #500
  - [x] multicenter-phantom 
  - [x] visual-working-memory #500
  - [x] Reusing-Neuro-Data #500
  - [x] cneuromod #502
  - [x] SIMON #500
- [ ] modify the DATS GUI editor to automatically save the date type in lower case except for 'CONP DATS fileset creation date', https://github.com/CONP-PCNO/conp-dats-editor/issues/15
- [x] modify the validator script and tests to enforce the convention #519

## Related issues 

#454

Issue #454, opened 2020-10-14
Harmonized values for our DATS datasets descriptors


jbpoline:
## Purpose
Suggest some harmonized values for the DATS model in the GUI, 

## Context
This would help find datasets with a specific values, make sure that these values are linked to a place with a definition, propose default values in the GUI

## Possible Implementation (optional)
Create a first json/md text files with most used / most recommended values

## List of fields to be harmonized:

- [x] units
- [x] file format #491 
- [x] date types #499 
- [x] privacy #532 
- [x] licenses #541
- [x] isAbout #533
- [x] keywords #543
- [x] data types #542
- [ ] dimensions

cmadjar:
@liamocn @PapillonMcGill could you discuss what the best format would be?

My guess would be the following:
- create a JSON file that could be parsed by the GUI to provide pre-established values for the fields
- would be nice for the fields to hint to the users what to enter when they start typing in the words

Thoughts? 

Thank you :)

cmadjar:
For the DataType, we could provide a list of terms from the interlex URI dashboard (https://neuinfo.org/interlex/dashboard). Is that the only source of terms to use there?

Suggestion to build list of terms on top of the NIDM-terms group. To be discussed today at the CONP dev.

PapillonMcGill:
No Interlex is not the only one. It is one if not the most comprehensive list of term for medical research. Easy to had new one if yours term is not there or definition do not mach perfectly your need.

PR #500, opened 2021-01-18
update datasets with correct date type format

cmadjar:
## Description

This is a continued attempt at harmonizing DATS values across datasets. The changes in this PR includes harmonization of the date types DATS field.

It updates almost all the discrepancies in datasets present under the conpdatasets organization. Note: DATS files for cneuromod and multicenter-phantom are not changed in this PR yet.

List of harmonized date types:
- CONP DATS fileset creation date
- end date
- last update date
- release date
- start date
- publication date

The general rule is that the date type should be all lower case except for CONP DATS fileset creation date since that value contains acronyms that should be in upper case.

## Related issues

#499 #454


emmetaobrien (approved):
This all looks fine to me.

PR #501, opened 2021-01-18
Fix date type values in DATS files of the missing two datasets

cmadjar:
## Description

This is a continued attempt at harmonizing DATS values across datasets. The changes in this PR include: 
- harmonization of the date types DATS field for multicenter-phantom and cneuromod
- change of git submodule to point to the conpdatasets repo instead of the user repo

List of harmonized date types:
- CONP DATS fileset creation date
- end date
- last update date
- release date
- start date
- publication date

The general rule is that the date type should be all lower case except for CONP DATS fileset creation date since that value contains acronyms that should be in upper case.

## Related issues 

#499 #454 #500


cmadjar:
I will be sending two separate PR: one for cneuromod and one for the multicenter phantom since this last one is more problematic...

Issue #495, opened 2021-01-12
Zenodo crawler does not go through all pages returned by the API response


cmadjar:
Current link used to query the Zenodo API: 

https://zenodo.org/api/records/?type=dataset&q=keywords:"canadian-open-neuroscience-platform"

At the bottom of the Zenodo response, there are the following information that can be used to navigate through pages a bit like the OSF crawler does:
```
"links": {
    "next": "https://zenodo.org/api/records/?sort=bestmatch&q=keywords%3A%22canadian-open-neuroscience-platform%22&type=dataset&page=2&size=10",
    "self": "https://zenodo.org/api/records/?sort=bestmatch&q=keywords%3A%22canadian-open-neuroscience-platform%22&type=dataset&page=1&size=10"
  }
```

PR #513, opened 2021-01-28
Loop through next pages to get the whole list of Zenodo datasets

cmadjar:
This fixes a bug in the Zenodo crawler where next pages of results to the Zenodo API query were not considered.

#495


glatard (approved):
lgtm!

Issue #515, opened 2021-01-28
List of datasets that are missing subjects value in extra property


cmadjar:
The following datasets are missing the subjects extra property value in their DATS.json files as flagged by #507:
- [ ] Multi-model_functionalization_of_disease-associated_PTEN_missense_mutations (waiting to hear back from Paul Pavlidis)
- [x] nhpatlas
- [ ] all openpain datasets (waiting to hear back from the openpain people for the number of subjects for their datasets)
- [ ] Systematic_phenomics_analysis_of_autism-associated_genes (waiting to hear back from Paul Pavlidis)

cmadjar:
@emmetaobrien the list above are what still needs to have the subjects extra property added to the dataset. I remember you contacted the openpain people and are still waiting for their answers.

Did you also contact the owners of nhpatlas, Multi-model_functionalization_of_disease-associated_PTEN_missense_mutations and Systematic_phenomics_analysis_of_autism-associated_genes? I don't remember.

Thanks!

emmetaobrien:
I have found the documentation for `nhpatlas` and updated that now.

emmetaobrien:
And I have mailed Paul Pavlidis to inquire after the subject counts for the other two datasets.

emmetaobrien:
Addressed in https://github.com/CONP-PCNO/conp-dataset/pull/516

PR #516, opened 2021-01-29
Updating datasets for number of subjects.

emmetaobrien:
This PR updates the `nhpatlas` dataset, six openpain datasets and two datasets from the Pavlidis lab, as specified in https://github.com/CONP-PCNO/conp-dataset/issues/515, to add numbers of subjects where meaningful, and a value of 0 in `extraProperties->subjects` where number of subjects is not meaningful.


cmadjar (approved):
LGTM

PR #519, opened 2021-02-10
Add validation of date type to the validator and the tests

cmadjar:
## Description

As an exercise towards DATS value harmonisation, the dates type have been harmonized to be all lower case with the exception of "CONP DATS fileset creation date".

This PR adds some functions to the validator and the automated tests to validate that the date type is indeed all lower case.

Note: since the DATS GUI is only suggesting to add dates to the dataset_schema.json and the primaryPublications section of the DATS file, the validation will only check those for now.

## Related issues 

#499

PR #520, opened 2021-02-10
Fix encoding of date type to be all lower case for the DATS file

cmadjar:
## Description

The date type of the DATS file created by the Zenodo crawler are `Date Created` and `Date Modified`. This changes those strings to comply with lower case convention, ergo `date created` and `date modified` so that the crawler creates valid DATS files.

## Related issues 

#519, #499


cmadjar:
@emmetaobrien if you could review and merge this one, that would be awesome. Thank you!

emmetaobrien (approved):
all good.

Issue #517, opened 2021-02-05
Recrawl preventad-open-bids and preventad-registered BIDS component


cmadjar:
The PREVENT-AD open BIDS dataset contained some empty 4D NIfTI files. Something went wrong with the mnc2nii conversion on Jul 27th for some obscure reason.

The dataset has been fixed on LORIS and need to be recrawled to fix the files on preventad-open-bids and the BIDS subdataset of preventad-registered

cmadjar:
Progress on the related tasks:
- [x] recrawl `preventad-open-bids` 
- [ ] recrawl `preventad-registered-images-in-BIDS`
- [ ] update `preventad-registered` with latest commit of `preventad-registered-images-in-BIDS`
- [ ] update conp-dataset with latest commit of `preventad-open-bids` and `preventad-registered`

Recrawling takes a lot of time because there are many files for these datasets...

PR #521, opened 2021-02-12
Update the PAD BIDS datasets and the datasets that needed date type updates in DATS

cmadjar:
## Description

This pulls the following changes in the datasets:
- preventad-open-bids and preventad-registered: fixes the URLs of some NIfTI files that were changed due to misconversion
- date type values harmonization for the datasets listed as problematic by the validator in #519

## Related issues 

#519, #517


emmetaobrien (approved):
LGTM

PR #528, opened 2021-02-24
[FIX] Docker workflow

mathdugre:
## Description
The current GitHub action for building and pushing the repo image to DockerHub is failing.
This is because the action uses a deprecated version of the `docker/build-push-action`.

This PR aims at updates the version for the docker action using the [current recommended approach](https://github.com/docker/build-push-action#git-context).

To be functional, this PR requires secrets for:

- [x] DOCKERHUB_USERNAME
- [x] DOCKERHUB_PASSWORD (should use a generated DockerHub Token)
- [x] DOCKERHUB_REPOSITORY

PR #529, opened 2021-02-24
[FIX] syntax error was introduced.

mathdugre:
It looks like I introduced a syntax error in one of the GitHub actions in #528.


cmadjar (approved):
oh, sorry. I missed that...

Issue #538, opened 2021-03-03
Update validator to check if there is at least one species entry in isAbout field


cmadjar:
This is a subtask of #533.

Add a check to the validator and tests to verify that there is at least one entry with an `identifier` and an `identifierSource` containing a value that starts with the string https://www.ncbi.nlm.nih.gov/Taxonomy.

Example structure of the isAbout field for a species entry:
```
   "isAbout": [
        {
            "identifier": {
                "identifier": "9483",
                "identifierSource":"https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=9483"
            },
            "name":"Callithrix jacchus"
        }
    ],
```

Note: this check should be skipped for DATS files from the crawlers (crawlers dataset always contain a `.conp-*-crawler.json` file at its route).

Issue #534, opened 2021-03-03
Validate values of the DATS privacy field


cmadjar:
Allowed values for the privacy field have been narrowed to the following 4 terms:
- open
- registered
- controlled
- private

Need to update the validator and the automated tests to validate for DATS `privacy` field values and ensure they match the convention above.

Note: the privacy field is not required in the DATS files but if present it would need to be one of the 4 terms above.

PR #545, opened 2021-03-04
Add privacy and species checks to DATS validator and tests

cmadjar:
## Description

This adds checks to the DATS validator for privacy and isAbout fields according to the description of the linked issues.

## Related issues

#534, #538


cmadjar:
@mathdugre looking at the Circle CI failure, I am surprised to not see any error for the isAbout field.

If I run `python scripts/dats_validator/validator.py --file=projects/preventad-open/DATS.json`, I get the following:
```
JSON schema validation passed.
Total required extra properties errors: 2
1 Validation error in PREVENT-AD open data: privacy - 'Open Access' is not allowed. Allowed value should be one of ['open', 'registered', 'controlled', 'private']. 
2 Validation error in PREVENT-AD open data: isAbout - There appears to be no species specified in isAbout. At least one species is required in the field and should follow the NCBI taxonomy. Valid example for a species:

    "isAbout": [
        {
            "identifier": {
                "identifier"      : "9606",
                "identifierSource": "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=9606"
            },
            "name"      : "Homo sapiens"
        }
    ],
```

So I know for sure that preventad-open should fail the privacy check and the isAbout check. However, in Circle CI I do not see any error regarding the isAbout check. I keep looking at the code but cannot see what I am missing. Any idea?

mathdugre:
@cmadjar, It looks like CircleCI reports an error about the *privacy* field then fails. The reason why the *isAbout* field error is not displayed is that the *privacy* field check is done beforehand and fails.

We should probably fix the privacy field in the dataset then check back on the test to make sure it works fine.
Another solution would be to split the PR to have (1) privacy test, and (2) isAbout test.
However, I think the first option is simplier.

cmadjar:
@mathdugre thank you for have checked! Indeed, we are fixing the privacy field of the datasets. Once they are all fixed, I will merge the latest commit of master into my branch and it will then be able to pick up the isAbout issue.

Thank you :)

cmadjar:
@emmetaobrien @mathdugre ready for review and hopefully a merge :)

cmadjar:
wouhou! Passing all 3 Circle CI checks! @emmetaobrien @mathdugre ready for you guys. After a month and a half of rebasing, I can't wait to have this merged ;-). Thanks!

mathdugre (approved):
LGTM :+1:

emmetaobrien (approved):
LGTM

Issue #549, opened 2021-03-09
Major test suite failure


mathdugre:
Currently, the testing suite is failing prior to testing datasets.
Failing due to: `./lint.sh: line 2: flake8: command not found`.

The expected behavior from #544 was to run flake8 prior to the tests.
However,  with the current caching system implement in the CircleCI workflow, adding dependencies to a *requirements.txt* also requires bumping the version of the cache.

I experienced the same issue several times during the initial development of the workflow. I think this goes to say that the current caching layer is error-prone and might need some rethinking.

## Environment (optional)

CircleCI master branch, [monitoring workflow](https://app.circleci.com/pipelines/github/CONP-PCNO/conp-dataset/2548/workflows/c6983257-9dc2-4ac8-96d5-a47836c6e9ad/jobs/5123/steps).

## Possible fix (optional)

I think that using the checksum of the aggregated *requirements.txt* in the cache key could solve the issue.

PR #550, opened 2021-03-09
Change to CircleCI dependencies caching

mathdugre:
## Description

This PR as for goal to reduce failures of the CircleCI pipeline after changing the dependencies of the project. See #549 for further detail.

The proposed solution is to merge all the *requirements.txt* in the repository and take unique values. Then calculate the checksum as a key to the cache. This way, the versioning of the cache would be done automatically as users change the dependencies.

Note: Having different **strict** versions of the same dependencies in two separates files will cause a failure of the pipelines.
e.g. `datalad==1.0.0` in */requirements.txt* and `datalad==1.1.1` in */scripts/requirements.txt*.

## Related issues (optional)

Issue #549


cmadjar:
@mathdugre failing circle CI

cmadjar:
@mathdugre could you merge master into your branch? The issue reported in Circle CI should be fixed now in master. Thanks!

cmadjar:
@mathdugre there are conflicts with the master branch, probably cause by the merge of your PR #551... Sorry.

I also looked at the Circle CI issue. This is one file that is always failing the tests, not because it cannot be downloaded but because the file takes time to be downloaded (probably because it was extracted from a zip file on Zenodo). Not sure what to do with this. I usually restart the Circle CI job hoping it will not try to download that file again but I am sure there are better ways to deal with this ;).

mathdugre:
@cmadjar, just fixed the merge conflict, and the tests are passing now :+1: .

I agree that the download timeout could be problematic when it comes to downloading archives from Zenodo. I opened #554 to further discuss this issue.

cmadjar (approved):
LGTM 👍 

Thank you @mathdugre!

PR #551, opened 2021-03-09
Add support for submodule test

mathdugre:
## Description

This PR proposes testing the download of submodule files in a dataset when the submodule is not a CONP dataset.
@cmadjar pointed out to me that some datasets might not be properly tested because they only contain submodules and no files annex at the root of the dataset; e.g. *preventad-registered*.

## Assumptions
A strong assumption made in this implementation is that submodules share the credentials and API endpoint with their parent dataset.

## Related issues (optional)

#340 Discussed the tradeoff of supporting submodules with different credentials and concluded that this is was not a desired feature at the time.


mathdugre:
@cmadjar, This PR currently seems to work on my CircleCI account. However, I currently don't have access to the *preventad-registered* dataset. Thus the download step was prompting the user to set a provider and set credentials.

We might want to verify that everything is working as expected when I get access to the dataset; or someone else.

tests/functions.py
```diff
@@ -263,3 +274,49 @@ def get_approx_ksmallests(dataset, filenames, k=4, sample_size=200):
...
+
+    parent_dataset_ids = set()
+    if "extraProperties" in dats:
+        for property_ in dats["extraProperties"]:
+            if property_["category"] == "parent_dataset_id":
+                parent_dataset_ids = {x["value"] for x in property_["values"]}
+                break
+
+    submodules = git.Repo(dataset).submodules
+    # Parent dataset should not be tested once; in their own dataset repository.
```
cmadjar, line 310:
```suggestion
    # Parent dataset should not be tested here but in their own dataset repository.
```

cmadjar (commented):
LGTM. See minor comment

cmadjar:
@mathdugre regarding the preventad-registered data, in theory Circle CI should pick that up as the environment variables were set in the config of this repo.

However, I see that Circle CI did not run on this PR either. I wonder why

cmadjar:
Note: I downloaded the entire dataset last week and it was working. Just FYI

mathdugre:
@cmadjar I think CircleCI does not rerun the workflow for a commit if it was run into another account/repo; which is odd. This is only from experience with it. There might be a setting to make sure it works properly.

I think it is possible to trigger the test from CircleCI or from a new commit. I will give a shot to the latter.

cmadjar:
@mathdugre could you merge master into your branch? The issue reported in Circle CI should be fixed now in master. Thanks!

cmadjar (approved):
LGTM

PR #555, opened 2021-03-12
Handle crawler git pull errors

joeyzhou98:
## Error
Currently, in `crawl.sh`, if the command `git pull --no-edit main master` fails, the touchfile is never removed and the crawler will never execute since it will always see that the touchfile is present until someone manually removes it

`git pull` failling:
image

Crawler never starting again:
image

## Fix
Only continue executing script if that command passes and handle creation and deletion of the touchfile in the shell script instead of letting the python script handle the deletion of the touchfile since the python script might not execute.

Also redirecting output of `git pull` to the logfile for easier debugging and allow for overriding the `BASEDIR` environment variable which is the path to the `conp-dataset` repo directory where the crawler does its job to allow other instances of the script to run in other locations

### DIFF
--- a/scripts/crawl.py
+++ b/scripts/crawl.py
@@ -93,9 +93,5 @@ if __name__ == "__main__":
         if repository.active_branch.name != "master":
             repository.git.checkout("master")
 
-        # Always clear .crawling touchfile
-        if ".crawling" in os.listdir("."):
-            os.remove(".crawling")
-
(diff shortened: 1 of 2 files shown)


cmadjar (approved):
aaaaah, that explains why sometimes I had to dig very deep to find out the proper log ;).

Thank you for the fix @joeyzhou98!

PR #559, opened 2021-03-17
adding most up to date version of BigBrain crawler to scripts directory. 

emmetaobrien:
This PR updates the scripts directory to include the most recent version of the BigBrain crawler.


scripts/BigBrain_crawler.py
```diff
@@ -0,0 +1,31 @@
+import os
+from ftplib import FTP
+
+
+host = 'bigbrain.loris.ca'
+directory = 'BigBrainRelease.2015/3D_Surfaces/Apr7_2016'
```
cmadjar, line 6: @emmetaobrien does that mean that for each BigBrain dataset, that needs to be changed? Maybe adding some documentation on how to use the script would be useful?

emmetaobrien:
This is a script @glatard sent me in Slack a little while ago which I am putting here for the sake of central availability.

cmadjar (approved):
LGTM. @glatard @mathdugre I am going to merge this but it would be good if the BigBrain crawling could be documented.

Issue #540, opened 2021-03-03
Fix wrong terms used in licenses field of the current DATS files


cmadjar:
The following identified values need to be changed in the DATS files:
- [x] CC-By Attribution 4.0 International => `CC BY-4.0`
- [x] Creative Commons Attribution 4.0 International (CC-BY 4.0)  => `CC BY-4.0`
- [x] CC-BY-4.0  => `CC BY-4.0`
- [x] BY-NC-SA => `CC BY-NC-SA`
- [x] Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International Public License => `CC BY-NC-SA`

PapillonMcGill:
4.0 (version) should go to the "version" tag
i.e.: 
"name":"CC BY",
"version":"4.0"

cmadjar:
oh, good catch! @zxenia, @emmetaobrien  I believe this will affect the NIDM term, no?

zxenia:
I'd still keep 4.0 in the name since it's the full name of the license here https://creativecommons.org/licenses/by/4.0/

cmadjar:
@PapillonMcGill @emmetaobrien I still see the following in the DATS files:

- `CC-By Attribution 4.0 International` should be replaced by `CC BY-4.0` for the following datasets
  - BigBrainHippoUnfold
  - Calgary_Preschool_MRI_Dataset
  - BigBrainMRICoreg
-  `Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International Public License` should be replaced by `CC BY-NC-SA` for the following datasets:
  - BigBrain
  - BigBrain_3DSurfaces
  - BigBrain_3DClassifiedVolumes

PR #558, opened 2021-03-17
updating 1000Genomes links

emmetaobrien:
This PR updates datafile links in the 1000Genomes subdataset, which were observed to be failing yesterday.

PR #560, opened 2021-03-18
fix licenses according to standard

PapillonMcGill:
## Description
fix licenses field according to new standard
added identifier when available

## Checklist

Mandatory files and elements:

Functional checks:

## Related issues (optional)
#540

### DIFF
--- a/scripts/conp_to_nidm_terms/README.md
+++ b/scripts/conp_to_nidm_terms/README.md
@@ -1,0 +1,18 @@
+### Scripts for creating report about distinct values and generating jsonld files for them
+
+
+- Create report about distinct values in each of the following properties: `privacy`, `types`, `licenses`, `isAbout`, `formats`, `keywords`. 
+By default generates report for all properties.
+
+<pre>python report_generator.py [--privacy=False --types=False --licenses=False --is_about= --formats=False --keywords=False --help]</pre>
+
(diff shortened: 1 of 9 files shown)


cmadjar:
@PapillonMcGill conflicts with the base branch.

The Circle CI error is due to the fact that the 1000GenomesProject does not include the latest change from Emmet in https://github.com/CONP-PCNO/conp-dataset/pull/558

cmadjar:
@PapillonMcGill there are still conflicts with the master branch

cmadjar (changes requested):
@PapillonMcGill In visual-working-memory, there is a typo in the license name (extra closing bracket that should not be there).

For the identifier field, I am noting that some of the licenses do not have identifier. Is that correct? Example CC BY-NC-SA

Finally, the indentation of the text in the change appears to be off for several datasets. Might be worth fixing those. Look at the github diff and you will see which datasets have an odd indentation in the license field of the DATS files.

Thank you!

PapillonMcGill:
@cmadjar For the identifier fields that are empty, it is because there is still no NIDM (or other) entry, waiting for Kzenia request to be approved.

Identation is due to the presence of `tab` in the text. I only used space and everything was fine inside my editor.
Do I replace all tab by 4 (or 2) spaces in those files. We should have a rule about no tab, this would prevent confusion.

PR #561, opened 2021-03-18
[FIX] Wrong file size reported when a download fails.

mathdugre:
## Description

This PR fixes the reported file size when testing the download of a file fails.
Currently, the test always show _inf_ . Moreover, this PR brings human format to the file size for easier readability.


cmadjar (approved):
👍

PR #567, opened 2021-03-25
Fix timeout of archived datasets in the testing suite

### LINKED ISSUE #554: Download timeout for Zenodo archived (Zip) datasets on CircleCI
When performing the download test of archived Zenodo datasets, the testing suite currently:
1. Download the whole archive to retrieve the file names.
2. Then tries to download individual files.

However, it seems that this strategy can be problematic as pointed out in #550, even for small archives.

To ensure archived Zenodo datasets are tested properly, we might need to find a more efficient approach to test the download of those files. 

I am opening this issue to discuss possible solutions.

mathdugre:
## Description
This PR aims at solving timeouts of archived datasets.
Two changes are introduced:
1. When getting files for download, those not part of an archive are prioritized.
2. Before downloading files, the test will now estimate a download time base on the file to download. i.e. Either normal file or archived ones. Note: The download time is capped at 10 minutes to avoid lengthy test.

## Related issues (optional)
fix #554

### DIFF
--- a/tests/functions.py
+++ b/tests/functions.py
@@ -227,19 +227,62 @@ def authenticate(dataset):
         )
 
 
-def get_filenames(dataset):
-    annex_list: str = git.Repo(dataset).git.annex("list")
-    filenames: List[str] = re.split(r"\n[_X]+\s", annex_list)[1:]
-    return filenames
+def get_filenames(dataset, *, minimum):
+    contains_archived_files = False
+    annex_list: str = iter(git.Repo(dataset).git.annex("list").split("\n"))
+    remotes = list()
+
+    # Retrieve remotes from the header.
+    for line in annex_list:
+        if re.match(r"^\|+$", line):
+            break
+        remotes.append(re.sub(r"^\|*", "", line))
+
+    if "datalad-archives" in remotes:
+        archived_files = list()
+        independent_files = list()
+        archive_index = remotes.index("datalad-archives")
+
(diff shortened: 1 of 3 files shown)


tests/functions.py
```diff
@@ -227,19 +227,62 @@ def authenticate(dataset):
...
+            contains_archived_files = True
+            filenames = archived_files + independent_files
 
+    else:
+        filenames = [x.split()[1] for x in annex_list]
+
+    return filenames, contains_archived_files
+
+
+def download_files(dataset, dataset_size, *, num=4):
```
cmadjar, line 267: @mathdugre shouldn't the `*` be in front of an argument? like `*arg`?

cmadjar (commented):
@mathdugre one tiny comment/question

tests/functions.py
mathdugre: @cmadjar the `*` signifies that further arguments are keyword-only. This enforce some arguments to be specified with their keyword.

For example:
`download_files("project/prevent-ad", 12345, 5)` would fail
 `download_files("project/prevent-ad", 12345, num=5)` is fine.


You can find more details about keyword-only arguments in [PEP 3102](https://www.python.org/dev/peps/pep-3102/).

cmadjar: Thank you @mathdugre !

cmadjar (approved):
LGTM

Issue #541, opened 2021-03-03
Harmonize DATS licenses field


cmadjar:
List of terms:
- CC BY
- CC BY-SA
- CC BY-NC
- CC BY-NC-SA
- CC BY-ND
- CC BY-NC-ND
- CC0
- ODbL
- ODC-By
- PDDL

**TODOs:**
- [x] find out which default value should be used when none are specified #539
- [x] fix datasets with wrong term for the license #540
- [x] crawlers default value is None for that field currently. Figure out whether we need to change that value to a default setting or keep it as None TO BE DISCUSSED
- [x] submit terms to NIDM #535

cmadjar:
@emmetaobrien @PapillonMcGill @zxenia if you have any ideas on how to incorporate this in the DATS validator. It might be tricky since users can enter URLs and other things in that field...

zxenia:
@cmadjar I don't think we need to add this in the dats validator because exactly what you said - we still want to allow users to use their own licenses and terms of use. When we prepare those terms we can add them in docs [here](https://github.com/CONP-PCNO/conp-documentation/blob/master/CONP_DATS_fields.md) as suggested licenses but leave a note that the users still can add their own. And once it's in the dats editor we minimize the probability that the same license is spelled differently.

Issue #533, opened 2021-03-03
Harmonize DATS isAbout field


cmadjar:
Values in this field should refer to biological entities. Species should always be included in this field. Other values can include diseases (Alzheimer's disease) or category (older adult, healthy...). 

Values in this field should be in lower case unless it is an acronym or know to required an upper case letter.

For the taxonomy, it should be using the terms from the NCBI taxonomy browser: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi

**TODOs**
- [x] fix the wrong terminology used for the `isAbout` field in the current DATS files #537
- [ ] modify the DATS GUI so that it reads the NCBI taxonomy to hint users for species entries [conp-dats-editor/issues/17](https://github.com/CONP-PCNO/conp-dats-editor/issues/17)
- [x] submit terms other than species to NIDM #535

Note: crawlers do not need to be updated for this field since the `isAbout` field cannot be automatically populated from Zenodo and OSF responses. This does not mean that some of the crawled dataset won't need to be updated if the users uploaded a DATS.json file. In which case, will need to contact the user so they can update their dataset.

cmadjar:
@emmetaobrien @PapillonMcGill @zxenia I am not sure to which extent we can add this field to the validator. For now, I did not mention the validator and tests in the description of this issue. Thoughts?

emmetaobrien:
We could require that under `isAbout` there must be an `identifier` with an `identifierSource` containing a value that starts with the string `https://www.ncbi.nlm.nih.gov/Taxonomy`; it's a bit clumsy but it's as good a way as I can see of confirming that the species is specified.

Issue #532, opened 2021-03-03
Harmonize DATS privacy field 


cmadjar:
**The allowed values for the privacy field should be one of the following terms:**
- open
- registered
- controlled
- private

**Proposed definitions for the 4 terms:**
- open: freely and publicly available
- registered: available to bona fide researchers/clinical care professionals only
- controlled: available to qualified researchers approved by a committee after review of their research proposal; also known as managed or restricted access
- private: available only to researchers of the project; also known as closed

**TODOs**
- [x]  add documentation about this field in [CONP DATS fields.md](https://github.com/CONP-PCNO/conp-documentation/blob/master/CONP_DATS_fields.md) - [conp-documentation/issues/59](https://github.com/CONP-PCNO/conp-documentation/issues/59)
- [x] update the validator and tests to allow only for those 4 terms in the privacy field #534 
- [x] review and correct the datasets for which the `privacy` field of the DATS file is incorrectly entered https://github.com/CONP-PCNO/conp-dataset/issues/536
- [x] check if crawlers need to be updated for this DATS field
- [x] modify the DATS GUI to allow for those 4 terms and include the definition [conp-dats-editor/issues/16](https://github.com/CONP-PCNO/conp-dats-editor/issues/16)
- [x] modify default displayed value on the portal for privacy from public to open [conp-portal/issues/414](https://github.com/CONP-PCNO/conp-portal/issues/414)
- [x] submit the terms to NIDM #535

cmadjar:
@emmetaobrien @PapillonMcGill @zxenia could you verify that the definition I propose above make sense? Thanks!

PapillonMcGill:
Yes, it make sense with what was discussed

emmetaobrien:
That all looks good to me.

cmadjar:
Crawlers do not need to be updated for this field. Yeah! One less thing to do :)

zxenia:
The definitions look good to me too.

PR #568, opened 2021-03-26
fix isAbout, types and licenses according to standard

PapillonMcGill:
## Description
fix multiple fields according to new standard list of term

## Checklist

Mandatory files and elements:

Functional checks:

## Related issues (optional)
#532 
#533
#541


cmadjar (changes requested):
@PapillonMcGill Thank you for this. 

I noticed a few things that need to be corrected:
- There is a lot of spacing issues introduced in the DATS files.
- SIMON dataset: in `types`, there is twice the value of `MRI` but when looking at the `identifier`, the second one should be `quality control subject`, not `MRI`
- I am confused, I see that in CHBMP, `MRI` is associated to a `valueIRI` but in SIMON, it is associated to an `identifier` section. Shouldn't they refer to the same? I see this inconsistency in several of the DATS files. Sometimes `valueIRI` is used, other times `identifier` fields.
- I don't think the `identifier` section of the `celltypes` DATS for type `transcriptomics` is correctly encoded. Also, why does the `transcriptomics` term have `identifier` associated to it under `types` and `valueIRI` in `keywords`? 
- `nhpatlas` dataset: major typos in types section for `transcriptomics` ;). Also, it has `identifier` associated to it in types but `valueIRI` for keywords, why?
- `mousebytes` dataset: `hasPart` of the DATS model needs to be updated for the `privacy` and `isAbout` fields

Thanks!

emmetaobrien:
With regard to the spacing issues, I should note that I have been consistently using tabs for indenting everywhere in DATS.json since the beginning of the project, so if @PapillonMcGill has edited files of mine using spaces for indenting that could cause oddities.

PapillonMcGill:
spacing issue is effectively caused by space vs tab. My editor is set to use 4 space as a tab as is required in LORIS.
I can replace every 4 space with a tab if you like but to prevent issue like that in the future, i suggest to opposite. replace all tab with 4 space (some DATS files use 2 space as identation). It is also easier to control no tab vs the right amount of spaces.

"valueIRI" is required for the keywords by the schema. When using "identifier", the file didn't validated .

I will check the dataset where error where found

cmadjar:
@PapillonMcGill there is something off with CHBMP. When I click on the diff on the frontend interface, I see no changes for that dataset.

The rest looks good. Thank you!

cmadjar (approved):

PR #577, opened 2021-04-08
Fix linting issue reported

mathdugre:
## Description

This PR aims to standardize the code style of the repository.
It solves the issues reported by the GitHub *linting* workflow .

## Currently addressed hooks

- [x] check-added-large-files
- [x] check-ast
- [x] check-json
- [x] check-merge-conflict
- [x] check-yaml
- [x] debug-statements
- [x] end-of-file-fixer
- [x] check-executables-have-shebangs
- [ ] mypy
- [x] pyupgrade
- [x] check-builtin-literals
- [x] check-docstring-first
- [x] pretty-format-json
- [x] trailing-whitespace
- [x] add-trailing-comma
- [x] black
- [x] blacken-docs
- [x] dockerfilelint
- [x] flake8
- [ ] pydocstyle
- [x] reorder-python-imports

PR #583, opened 2021-04-09
[MODIFY] Bump Python version to 3.9 in CircleCI.

mathdugre:
## Description

This PR aims at bumping the Python version to 3.9 in CircleCI.

It was discussed in #577 that the python typing module went through a lot of changes in the recent version.
Upgrading the Python version used for the development of the test suite would result in better readability and maintenance; see [PEP 585](https://www.python.org/dev/peps/pep-0585/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 aliases:
   - &docker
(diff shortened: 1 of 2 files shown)


cmadjar (approved):
👍

PR #586, opened 2021-04-13
Modify date type encoded by the OSF crawler to match DATS harmonization work

cmadjar:
## Description

This modifies how the date types are encoded to match with the DATS harmonization work that has been done. This was modified for the Zenodo crawler but not for the OSF crawler for some unclear reasons. Probably was just missed.

Related to #520 that corrected that for the Zenodo Crawler.


emmetaobrien (approved):
LGTM

PR #605, opened 2021-04-30
Crawlers check for licenses if none was provided

joeyzhou98:
## Description
From discussion with Emmet, we want to make sure crawled datasets have their license field filled in in their DATS.json, however, some datasets on OSF and Zenodo do not specify a license on the platform and will instead add a license file somewhere in the dataset.

This PR will add the ability for the crawlers to check and fill in the license field in the DATS if no license was provided from the platform, by gathering all license files and verifying if any of our supported license codes are contained within them and will fill in the DATS according to what code it finds inside the license file. An example of this feature in action would be in #609 and more specifically, https://github.com/conp-bot/conp-dataset-Synthetic-Animated-Mouse-SAM-University-of-British-Columbia-Datasets-and-3D-models/compare/38bb037796a7d2c89d77a381b7d42b7d46b2deb0...3a7c34ef24a81fb6882794d9fe73812b915ac3b1#diff-867e909ca880a8f815013d579cd988d98f8bc458db7ace379116498186ac03d9 where the license field was filled in automatically.

The crawler will also, by default, not annex files containing the string `license` in order to be able to see those potential license files on github as seen on https://github.com/conp-bot/conp-dataset-Synthetic-Animated-Mouse-SAM-University-of-British-Columbia-Datasets-and-3D-models/compare/38bb037796a7d2c89d77a381b7d42b7d46b2deb0...3a7c34ef24a81fb6882794d9fe73812b915ac3b1#diff-aad429a7482d1bae04b4e8455dbb9ebbdd2d784239e8ffc5af1612986df03daa.

I've also cleaned up the code by putting all constants in their own file.


emmetaobrien:
I'm sorry, I don't think I communicated clearly what I was asking for here; this isn't quite the desired behaviour.

If a license file can be found in the dataset, the intended logic is not to take the first line, which is not guaranteed to contain the information we want.

The desired behaviour when a license file is found is:
1) obtain the list of known valid license codes (as currently attached to the license field in the DATS file editing GUI at https://portal.conp.ca/share; I am not sure where this is stored and it may be hardwired into the page, I believe hardwiring it into the crawler would not be a problem as I do not anticipate this changing often or in major ways.)
2) for each valid license code, search the license file for it; if the license file is, as we have seen happens, just a copy of a known license, the code will occur during it at some point.
3) populate the DATS license field with the code when found.

If the license cannot be identified from the file in this way, then it is presumably a custom license of some sort, which probably wants wider discussion as to how to handle; if we are moving towards a policy of asking users to generate their DATS.json file with our GUI before adding data to a repository we crawl, this may become less of an issue generally.

joeyzhou98:
@emmetaobrien sorry I should have gathered the feature requirements more in details, my fault. I will implement it as requested above

joeyzhou98:
@emmetaobrien I have updated the description to reflect all the features implemented in this PR. Please let me know if there is anything else you want to have added or questions

scripts/Crawlers/constants.py
```diff
@@ -0,0 +1,60 @@
...
+    "hasPart",
+    "acknowledges",
+]
+MODALITIES = {
+    "fMRI": ["bold", "func", "cbv"],
+    "MRI": ["T1", "T2", "FLAIR", "FLASH", "PD", "angio", "anat", "mask"],
+    "diffusion": ["dwi", "dti", "sbref"],
+    "meg": ["meg"],
+    "intracranial eeg": ["ieeg"],
+    "eeg": ["eeg"],
```
cmadjar, line 57:
```suggestion
    "MEG": ["meg"],
    "intracranial EEG": ["ieeg"],
    "EEG": ["eeg"],
```

in theory, MEG and EEG should be all upper case as they are acronyms like fMRI and MRI

cmadjar (approved):
LGTM 👍

cmadjar:
@emmetaobrien could you do a last round of review on this (including the description that @joeyzhou98 updated). Just to be sure I did not miss anything.

@joeyzhou98 love the refactoring with the constant variables 👍

emmetaobrien:
This seems solid to me, thank you @joeyzhou98 .

PR #611, opened 2021-05-06
Addition of the BigBrain ROIs dataset

mathdugre:
## Description

This PR adds the BigBrain ROIs dataset crawled from the [loris database](https://bigbrain.loris.ca/main.php).

## Checklist

Mandatory files and elements:

Functional checks:


.gitmodules
```diff
@@ -124,6 +124,9 @@
 [submodule "projects/multicenter-phantom"]
 	path = projects/multicenter-phantom
 	url = https://github.com/conpdatasets/multicenter-phantom.git
+[submodule "projects/Bigbrain_3DROIS"]
```
cmadjar, line 127: In the git submodule, the s after ROI is upper case but in the actual directory it is lower case.

.gitmodules
```diff
@@ -124,6 +124,9 @@
...
+	path = projects/Bigbrain_3DROIS
```
cmadjar, line 128:
```suggestion
	path = projects/Bigbrain_3DROIS
```

but the second file in the file diff of Github is `projects/BigBrain_3DROIs` (lower case `s` vs upper case `S`)

.gitmodules
```diff
@@ -124,6 +124,9 @@
...
+	url = https://github.com/mathdugre/conp-dataset-Bigbrain-3DROIS.git
```
cmadjar, line 129: Instead of having the dataset in your username, could you publish the DataLad dataset to https://github.com/conpdatasets/BigBrain_3DROIs instead and link that repo in the submodule?

We try avoiding to have datasets hosted on our own accounts as much as possible.

cmadjar (changes requested):
In the files diff of the PR, I cannot click on the submodule. It could be related to the inconsistencies in the paths specified in submodule and the real path?

.gitmodules
cmadjar: I just gave you write access to the repo so you can publish directly to it via DataLad.

Maybe the reason the repo cannot be forked is because it is empty?

mathdugre: I brought the change to the submodule url.

cmadjar (approved):
LGTM 👍 

Thank you @mathdugre :)

Issue #623, opened 2021-05-19
Modify the tests of DATS file so that new prospective dataset sent via Zenodo, OSF or FRDR have a complete validation of the DATS file


cmadjar:
## Purpose

This has been discussed during the CONP dev meeting of May 19th, 2021.

Need to modify the automated tests so that new datasets crawled from Zenodo or OSF can have their DATS file fully validated against the CONP requirements.

In order to not fail all of the retrospective datasets, need to add a list of those retrospective datasets for which the DATS file will not be fully validated as the people of those datasets will be contacted later on. However, need to provide a list of datasets that fails the validation in the meantime.

cmadjar:
List of datasets DATS that fail the DATS validator:

- [ ] CFMM_7T__MP2RAGE_T1_mapping
```
JSON schema validation passed.
Total required extra properties errors: 3
1 Validation error in CFMM-7T: MP2RAGE T1 mapping: extraProperties.category.subjects is required but not found.
2 Validation error in CFMM-7T: MP2RAGE T1 mapping: extraProperties.category.CONP_status is required but not found.
3 Validation error in CFMM-7T: MP2RAGE T1 mapping: distributions.formats - 'formats' property is missing under distributions. Please add the 'formats' property to 'distributions'.
```
- [ ] Calgary_Preschool_MRI_Dataset
```
JSON schema validation passed.
Total required extra properties errors: 3
1 Validation error in Calgary Preschool MRI Dataset: extraProperties.category.subjects is required but not found.
2 Validation error in Calgary Preschool MRI Dataset: extraProperties.category.CONP_status is required but not found.
3 Validation error in Calgary Preschool MRI Dataset: distributions.formats - 'formats' property is missing under distributions. Please add the 'formats' property to 'distributions'.
```
- [ ] Comparing_Perturbation_Modes_for_Evaluating_Instabilities_in_Neuroimaging__Processed_NKI_RS_Subset__08_2019_
```
JSON schema validation passed.
Total required extra properties errors: 1
1 Validation error in Comparing Perturbation Modes for Evaluating Instabilities in Neuroimaging: Processed NKI-RS Subset (08/2019): extraProperties.category.subjects is required but not found.
```
- [ ] Intracellular_Recordings_of_Murine_Neocortical_Neurons
```
JSON schema validation passed.
Total required extra properties errors: 3
1 Validation error in Intracellular Recordings of Murine Neocortical Neurons: extraProperties.category.subjects is required but not found.
[...]

PR #624, opened 2021-05-19
Modify the tests to check all properties even from the Zenodo and OSF crawled datasets

cmadjar:
## Description

This resolves the issue of newly crawled datasets always passing the test even when the DATS.json file was automatically generated, since some required fields cannot be generated automatically by the crawlers.

From now on, ALL datasets need to have a fully validated DATS.json file. We will start by flagging the new datasets to be added if they do not have a valid DATS.json file. 

Users that already have their data uploaded to Zenodo or OSF on the portal will be later contacted to ask them the provide a proper DATS file and a more complete README file. Since this process might take a while, in the meantime, we will skip the extra validation checks for those datasets which are listed in a constant file. That file will be updated at the same time the user updates their DATS and README files.

When we get to the stage where all datasets have been updated, we will remove the constant always perform the extra checks.

## Related issues 

#623


cmadjar:
@emmetaobrien ready for review.

emmetaobrien (approved):
This all looks like it makes sense to me.

Issue #552, opened 2021-03-10
refseq_Homo_sapiens


cmadjar:
* Test build: 
https://app.circleci.com/pipelines/github/CONP-PCNO/conp-dataset/2556/workflows/7736ed3c-f4c5-4ab9-b7cc-8a4390c424d8/jobs/5142
https://circleci.com/gh/CONP-PCNO/conp-dataset/5144?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://circleci.com/gh/CONP-PCNO/conp-dataset/5139?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

* Standard error:
```
projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz
verification of content failed
```

Manual download appears to be downloading the file but we get the following error:
```
(datalad_python3) refseq_Homo_sapiens $ datalad get Homo_sapiens.gff.gz 
[WARNING] Running get resulted in stderr output: git-annex: get: 1 failed       
[ERROR  ] verification of content failed [get(/Users/cmadjar/tmp_con/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz)] 
get(error): /Users/cmadjar/tmp_con/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz (file) [verification of content failed]
```
Possible that the file changed as the reference human genome sequence gets updated regularly.

Issue #474, opened 2020-12-04
refseq_Homo_sapiens


cmadjar:
Cannot download the file Homo_sapiens.gff.gz. 

DataLad message:
```
[ERROR  ] verification of content failed [get(/data-scratch1/for_CONP_portal/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz)] 
get(error): Homo_sapiens.gff.gz (file) [verification of content failed]
```

cmadjar:
@emmetaobrien if you could investigate as to why this particular files, that would be awesome. Thank you!

Issue #402, opened 2020-08-26
refseq_refseq_Homo_sapiens


glatard:
* Test build: 
https://app.circleci.com/pipelines/github/CONP-PCNO/conp-dataset/947/workflows/3ba5e87d-9216-42cc-a45f-69c189f10b10/jobs/1901

* Standard error:
```
________ TestDataset.test_download[projects/refseq/refseq_Homo_sapiens] ________
projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz
verification of content failed
----------------------------- Captured stderr call -----------------------------
[WARNING] Running get resulted in stderr output: git-annex: get: 1 failed
 
[ERROR] verification of content failed [get(/home/circleci/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz)] 
```

* Help requested from:
@emmetaobrien

emmetaobrien:
This is fixed as of https://github.com/CONP-PCNO/conp-dataset/pull/404.

Issue #323, opened 2020-06-03
refseq_Homo_sapiens fails data download


glatard:
* Test build: https://app.circleci.com/pipelines/github/CONP-PCNO/conp-dataset/245/workflows/6b854ba7-56f6-4894-848b-f869de4ee4fe/jobs/459/parallel-runs/0?filterBy=FAILED
* Standard error:
```
________ TestDataset.test_download[projects/refseq/refseq_Homo_sapiens] ________
 
projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz
 
verification of content failed
 
----------------------------- Captured stderr call -----------------------------
 
[WARNING] Running get resulted in stderr output: git-annex: get: 1 failed
 
 
 
[ERROR] verification of content failed [get(/home/circleci/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz)] 
 
------------------------------ Captured log call -------------------------------
 
WARNING  datalad.annex:annexrepo.py:2268 Running get resulted in stderr output: git-annex: get: 1 failed
 

 
ERROR    datalad.distribution.get:utils.py:563 verification of content failed [get(/home/circleci/conp-dataset/projects/refseq/refseq_Homo_sapiens/Homo_sapiens.gff.gz)]
 ```
* Help requested from: @emmetaobrien

PR #627, opened 2021-05-20
Zenodo crawler private dataset refactor

joeyzhou98:
## Description
We used to crawl private datasets on Zenodo by appending the dataset token in the URL which leaks it to the public in the git annex history. Now we use bearer token to authenticate via `DATALAD_ZENODO_token` environment variable which is more secure since it doesn't leak the token in the commit history

### DIFF
--- a/scripts/Crawlers/ZenodoCrawler.py
+++ b/scripts/Crawlers/ZenodoCrawler.py
@@ -1,28 +1,23 @@
 import datetime
 import json
 import os
+from typing import Callable
 
 import html2markdown
 import humanize
 import requests
+from datalad.distribution.dataset import Dataset
 
 from scripts.Crawlers.BaseCrawler import BaseCrawler
 
 
-def _get_unlock_script():
-    with open(os.path.join("scripts", "unlock.py")) as f:
-        return f.read()
-
-
-def _create_zenodo_tracker(path, dataset, private_files, restricted):
+def _create_zenodo_tracker(path, dataset):
     with open(path, "w") as f:
         data = {
             "zenodo": {
                 "concept_doi": dataset["concept_doi"],
                 "version": dataset["latest_version"],
             },
-            "private_files": private_files,
-            "restricted": restricted,
             "title": dataset["title"],
         }
         json.dump(data, f, indent=4)
@@ -32,7 +27,6 @@ class ZenodoCrawler(BaseCrawler):
     def __init__(self, github_token, config_path, verbose, force, no_pr):
         super().__init__(github_token, config_path, verbose, force, no_pr)
         self.zenodo_tokens = self._get_tokens()
-        self.unlock_script = _get_unlock_script()
 
     def _get_tokens(self):
         if os.path.isfile(self.config_path):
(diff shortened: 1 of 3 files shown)


cmadjar:
@joeyzhou98 looks like the tests are broken in this PR and I am not sure to understand why the openpain datasets show up as broken here as well (I see the same result in the PERFORM PR BTW).  

Currently, the only two datasets that are broken are: PERFORM and refseq_Homo_sapiens.

joeyzhou98:
@cmadjar I don't think the problem comes from this PR, but my guess is that someone merged something in CONP-PCNO:conp-dataset that broke those tests because this PR doesn't change any of the PERFORM nor refseq_Homo_sapiens submodules. Maybe @mathdugre knows more

joeyzhou98:
Oh wait, I know why the PERFORM dataset is failing. I had to recrawl the PERFORM dataset for security issues so I deleted the old repo and the test is trying to install the old one. This also explains why the PERFORM tests aren't failing for https://github.com/CONP-PCNO/conp-dataset/pull/629 but are failing for this PR. Notice the url is different for the PERFORM dataset submodule: https://github.com/CONP-PCNO/conp-dataset/pull/629/files#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584L28-R150. Now as for refseq, that is outside my control. The PERFORM tests should work once https://github.com/CONP-PCNO/conp-dataset/pull/629 is merged

mathdugre:
> I deleted the old repo and the test is trying to install the old one.

I agree that this is what is causing the test failure for both the *main* branch and this PR.

As for the `refseq_Homo_sapiens` dataset, this issue seems  to be recurrent; around every 3 months (#552, #474, #402, #323). I'm not sure how this used to be solve, maybe @emmetaobrien has some insight on this.
Note: This should be fixed in another PR though

cmadjar:
@mathdugre @joeyzhou98 Indeed, @emmetaobrien will fix the issue for the refseq dataset, see https://github.com/CONP-PCNO/conp-dataset/issues/638 

Thank you for the clarification for the PERFORM dataset. Makes sense.

scripts/Crawlers/ZenodoCrawler.py
```diff
@@ -412,3 +362,53 @@ def get_readme_content(self, dataset):
...
+        with open(zenodo_config_path, "w") as f:
+            f.write(
+                """[provider:ZENODO]
+url_re = .*zenodo\\.org.*
+authentication_type = bearer_token
+credential = ZENODO
+
+[credential:ZENODO]
+# If known, specify URL or email to how/where to request credentials
+# url = ???
```
cmadjar, line 395: Shouldn't `url` be set here?

joeyzhou98: Nope, it is only needed when requesting credentials

cmadjar: right! Thank you :)

cmadjar (approved):
LGTM! Thank you @joeyzhou98

PR #621, opened 2021-05-12
Addition of the BigBrain derivatives

mathdugre:
## Description

This PR aims to integrate missing BigBrain derivatives in CONP.

Below is a list of all BigBrain derivatives found on their FTP server for the 2015 release:
- [ ] A3D
- [x] ~3D_Surfaces~ (Already integrated)
- [x] ~3D_ROIs~ (Already integrated)
- [ ] MRISIM
- [ ] 3D_MRI
- [ ] BigBrainWarp_Support
- [x] ~3D_Classified_Volumes~ (Already integrated)
- [ ] Hippocampus_Segmentation
- [x] ~3D_Volumes~ (Already integrated)
- [x] ~3D_Blocks~ (Already integrated)
- [ ] Surface_Parcellations
- [ ] Layer_Segmentation
- [ ] 2D_Final_Sections

## Checklist

Mandatory files and elements:
- [ ] A `README.md` file, at the root of the dataset
- [ ] A `DATS.json` file, at the root of the dataset
- [ ] If configuration is required (for instance to enable a special remote), a `config.sh` script at the root of the dataset
- [ ] A DOI (see instructions in [contribution guide](https://github.com/CONP-PCNO/conp-dataset/blob/master/.github/CONTRIBUTING.md), and corresponding badge in `README.md`

Functional checks:
- [ ] Dataset can be installed using DataLad, recursively if it has sub-datasets
- [ ] Every data file has a URL
- [ ] Every data file can be retrieved or requires authentication
- [ ] `DATS.json` is a valid DATs model
- [ ] If dataset is derived data, raw data is a sub-dataset

PR #635, opened 2021-05-25
BigBrain_Surface_Parcellations: [ADD] crawled dataset.

mathdugre:
## Description

This PR integrate one of the missing bigbrain derivatives listed in #621 .

## Checklist

Mandatory files and elements:

Functional checks:
- [x] If dataset is derived data, raw data is a sub-dataset

## Related issues (optional)

#621

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -148,3 +148,6 @@
 [submodule "projects/openpain/thermal"]
 	path = projects/openpain/thermal
 	url = https://github.com/conpdatasets/openpain-thermal.git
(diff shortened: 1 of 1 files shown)


cmadjar (changes requested):
@mathdugre I added the the DATS and README to that dataset. I notice that there is no bigbrain-datalad submodule in this dataset but since it is derived from it, the bigbrain-datalad github submodule should be in the datasets as well.

cmadjar (approved):
LGTM. Thank you!

PR #637, opened 2021-05-27
[FIX] [REFACTOR] FTP crawler

mathdugre:
## Description

When there was an "[Err 104] Connection reset by peer" the script was hanging.
Also improved the overall performance and use git api instead of OS call.

### DIFF
--- a/scripts/datalad_crawlers/ftp_crawler.py
+++ b/scripts/datalad_crawlers/ftp_crawler.py
@@ -52,7 +52,7 @@ def parse_args():
     parser.add_argument("host", type=str, help="URL of the FTP host.")
     parser.add_argument("directory", type=str, help="Directory path to the dataset.")
     parser.add_argument(
-        "sub_directory", type=str, default="", help="Subdirectory to crawl."
+        "sub_directory", nargs="?", type=str, default="", help="Subdirectory to crawl."
     )
 
     return parser.parse_args()
--- a/scripts/datalad_crawlers/ftp_crawler.py
+++ b/scripts/datalad_crawlers/ftp_crawler.py
@@ -2,24 +2,37 @@ import argparse
 import os
 from ftplib import FTP
 
+import git
+from tqdm import tqdm
 
-def crawl(host, root, subdir):
+
+def crawl(host, root, subdir, *, ftp):
     """
     Recursively prints the URLs of the files under dir and adds them to git-annex
     """
-    old_dir = ftp.pwd()
-    ftp.cwd(os.path.join("/", root, subdir))
-    for entry in ftp.mlsd():
-        file_name, metadata = entry
-        if metadata["type"] == "dir":
-            print(f"mkdir -p {file_name}")
-            crawl(host, root, os.path.join(subdir, file_name))
-        elif metadata["type"] == "file":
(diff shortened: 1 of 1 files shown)


scripts/datalad_crawlers/ftp_crawler.py
```diff
@@ -2,24 +2,37 @@
...
+                elif metadata.get("type") == "file":
+                    git.Repo().git.annex(
+                        "addurl",
+                        f"ftp://{host}/" + os.path.join(cwd, filename),
+                        file=filepath,
+                    )
+                break
+
+            except Exception as e:
+                print(f"WARNING: Connection restarted on file: {filepath}: {e}")
```
cmadjar, line 33: I see that the FTP connection is reset in here but if the connection failed on the file, doesn't it mean that this file is skipped and it goes to the next filename?

cmadjar:
@mathdugre One question/comment above. Thank you for submitting the changes to the script!!

mathdugre:
@cmadjar , we might want to wait before merging the PR. I'm still experiencing some issue when crawling a large dataset; if there's an internet problem on the user side, the user will have to restart  the crawling from scratch.
I'm currently implementing the feature to crawl only a specific file/sub-directory of the dataset.

mathdugre:
@cmadjar, I implemented the feature to crawl subdirectories. The PR will be ready for review again.

scripts/datalad_crawlers/ftp_crawler.py
cmadjar: @mathdugre did you answer my question above? Sorry if I don't remember. I don't see anything in the comments of the PR as far as I can tell.

cmadjar:
@mathdugre see question above I had: https://github.com/CONP-PCNO/conp-dataset/pull/637#discussion_r646890211

cmadjar:
@mathdugre you will also need to merge master into your branch to fix some of the Circle CI errors.

mathdugre:
> @mathdugre see question above I had: [#637 (comment)](https://github.com/CONP-PCNO/conp-dataset/pull/637#discussion_r646890211)

Not sure what happened with my answer, sorry for that.

The crawler will only proceed to the next file it succeed to download the current one. That is, it will retry the same file again and again until it is done downloading the file/folder; see line 19 and 30.
Do you think it would be good to have a maximum amount of tries? If so, should the crawler stop after a maximum retry failure? Or proceed with the next file and log the error in a file for the user to see?

cmadjar:
@mathdugre I guess the WARNING will keep showing as long as the file cannot be downloaded. If so, I guess the user can decide to cancel whenever they want. Maybe it is best to let it run, just in case the connection loss is momentary?

Would that make sense leaving it running and letting the user stop the run if they think it is taking too long? If not, I would just end the crawling with the error message displayed in the terminal.

For sure, I would not proceed to the next file as there are big chances that the failure will be missed in case someone forgets to look at the log.

mathdugre:
@cmadjar, I agree that we should not skip a file that was not downloaded, as the user can miss the warning.
I think having the script run until it can download the file should be fine. If some issues arise with time, we can change it to a max attempts before exiting with error.

cmadjar (approved):
LGTM. Thank you @mathdugre !

PR #640, opened 2021-06-01
Add DATS JSONLD annotator

surchs:
## Description
a tool to turn a DATS.json file into a DATS.jsonld file with the appropriate @type declarations and the specific @context mapping needed for this instance.
* tests for the dats annotator
* Documentation for the dats jsonld annotator
* a readme
* Added submodules for the CONP forks of DATS context and DATS schema

The idea is that this tool can be used to to convert our DATS.json files to DATS.jsonld files that contain their own @context and have their @type annotations to be then taken into Nexus.

### DIFF
--- a/scripts/dats_jsonld_annotator/README.md
+++ b/scripts/dats_jsonld_annotator/README.md
@@ -1,0 +1,29 @@
+# DATS JSONLD annotator
+
+
+This is a command line tool to turn a [DATS.json file](https://github.com/datatagsuite) into a DATS.jsonld file with the appropriate `@type` declarations and the 
+[schema.org](https://schema.org/) `@context` mapping needed for this specific file. To do this, the tool relies on the [CONP specific DATS-schema files](https://github.com/CONP-PCNO/schema) and 
+the [CONP specific DATS-SDO-context files](https://github.com/CONP-PCNO/context).
+
+## Usage
+
+There are two main options:
+
+**Option 1**:
+
+Annotate all files called `DATS.json` in the subdirectories of a directory.
+
+```python
+python annotator.py /path/to/project/folder --out /path/to/output/directory
+```
+
+The `--out` parameter is optional, if it is not specified, the files will be generated in the 
+same directory as the original DATS.jsonld.
+
+**Option 2**:
+Annotate a single DATS file 
+```python
+python annotator.py /path/to/a/DATS_file.json --out /outputpath
+```
+
+The `--out` parameter can either be a file name or a directory.
--- a/scripts/dats_jsonld_annotator/README.md
+++ b/scripts/dats_jsonld_annotator/README.md
@@ -1,8 +1,8 @@
 # DATS JSONLD annotator
 
 
-This is a command line tool to turn a [DATS.json file](https://github.com/datatagsuite) into a DATS.jsonld file with the appropriate `@type` declarations and the 
-[schema.org](https://schema.org/) `@context` mapping needed for this specific file. To do this, the tool relies on the [CONP specific DATS-schema files](https://github.com/CONP-PCNO/schema) and 
+This is a command line tool to turn a [DATS.json file](https://github.com/datatagsuite) into a DATS.jsonld file with the appropriate `@type` declarations and the
+[schema.org](https://schema.org/) `@context` mapping needed for this specific file. To do this, the tool relies on the [CONP specific DATS-schema files](https://github.com/CONP-PCNO/schema) and
 the [CONP specific DATS-SDO-context files](https://github.com/CONP-PCNO/context).
 
 ## Usage
@@ -13,16 +13,16 @@ There are two main options:
 
 Annotate all files called `DATS.json` in the subdirectories of a directory.
 
-```python
+```shell
 python annotator.py /path/to/project/folder --out /path/to/output/directory
 ```
 
-The `--out` parameter is optional, if it is not specified, the files will be generated in the 
+The `--out` parameter is optional, if it is not specified, the files will be generated in the
 same directory as the original DATS.jsonld.
 
 **Option 2**:
-Annotate a single DATS file 
-```python
+Annotate a single DATS file
+```shell
 python annotator.py /path/to/a/DATS_file.json --out /outputpath
 ```
 
--- a/scripts/dats_jsonld_annotator/README.md
+++ b/scripts/dats_jsonld_annotator/README.md
@@ -18,7 +18,7 @@ python annotator.py /path/to/project/folder --out /path/to/output/directory
 ```
 
 The `--out` parameter is optional, if it is not specified, the files will be generated in the
-same directory as the original DATS.jsonld.
+same directory as the original DATS.json.
 
 **Option 2**:
 Annotate a single DATS file
--- a/scripts/dats_jsonld_annotator/annotator.py
+++ b/scripts/dats_jsonld_annotator/annotator.py
@@ -202,7 +202,7 @@ def gen_jsonld_outpath(dats_json_f, out_path):
         dats_jsonld_f = out_path / out_name
     else:
         raise Exception(
-            f"{out_path = } for {dats_json_f.resolve} is not a path to a file or directory. "
+            f"{out_path = } for {dats_json_f.resolve()} is not a path to a file or directory. "
             f"I don't know where to store the output. "
             f"Please provide a valid path with the --out flag."
         )
--- a/scripts/dats_jsonld_annotator/annotator.py
+++ b/scripts/dats_jsonld_annotator/annotator.py
@@ -1,9 +1,10 @@
-import json
-import sys
-import time
 import argparse
+import json
 import logging
 import pathlib as pal
+import sys
+import time
+
 import jsonschema as jss
 
 
@@ -103,7 +104,9 @@ def find_context(schema_id, context_dir):
     return context
 
 
-def annotate_dats_object(json_object, schema, specific_context, context_dir=CONTEXT_DIR):
+def annotate_dats_object(
+    json_object, schema, specific_context, context_dir=CONTEXT_DIR
+):
     """
     This function recursively traverses a DATS instance and generates two things:
 
@@ -255,7 +258,7 @@ def main(cli_args):
         type=pal.Path,
         help="""
                 If this is a path to a DATS file, then only this DATS file is annotated.
-                If this is a path to a directory, then each subdirectory of this directory is expected 
+                If this is a path to a directory, then each subdirectory of this directory is expected
                 to contain a DATS file called DATS.json. Each of these files will then be annotated iteratively
                 """,
     )
--- a/scripts/dats_jsonld_annotator/annotator.py
+++ b/scripts/dats_jsonld_annotator/annotator.py
@@ -289,7 +289,9 @@ def main(cli_args):
     elif args.dats_path.is_dir():
         files_to_convert = list(args.dats_path.glob("*/DATS.json"))
         if not args.out.is_dir():
-            logger.warning(f"The {args.out.resolve()} folder will be created and JSONLD files will be saved in it.")
+            logger.warning(
+                f"The {args.out.resolve()} folder will be created and JSONLD files will be saved in it."
+            )
             args.out.mkdir()
         if files_to_convert is None:
             logger.error(
--- a/scripts/dats_jsonld_annotator/annotator.py
+++ b/scripts/dats_jsonld_annotator/annotator.py
@@ -1,0 +1,320 @@
+import json
+import sys
+import time
+import argparse
+import logging
+import pathlib as pal
+import jsonschema as jss
+
+
+CONTEXT_DIR = (pal.Path(__file__).parent / "context" / "sdo").resolve()
+SCHEMA_DIR = (pal.Path(__file__).parent / "schema").resolve()
+logger = logging.getLogger("DATS annotator")
+
+
+def find_schema(parent_schema, term, json_object):
+    """
+    This function finds the appropriate JSON schema for a term in a DATS Json object. To do
+    so, it relies on the JSON schema of the supplied JSON object.
+
+    :param parent_schema: The DATS JSON schema that corresponds to the json_object parameter
+    :param term: the string term for which the schema shall be found
+    :param json_object:  the JSON object that contains the term and to which the parent_schema corresponds
+    :return: the DATS JSON schema for the term
+    """
+    ps = parent_schema["properties"]
+    if term not in ps.keys():
+        # TODO: Decide whether to raise an Exception here
+        # The only way this could happen on a valid dataset is if someone has specified additional properties
+        # We'd like to discourage this
+        logger.debug(
+            f'I cannot find {term = } in the parent schema: {parent_schema["id"]}.\n{json_object = }'
+        )
+        return None
+
+    # Set up the schema resolver
+    resolver = jss.RefResolver(base_uri=parent_schema["id"], referrer=None)
+
+    # See if "items" is in the schema entry
+    search_dict = ps[term]
+    if "items" in search_dict.keys():
+        search_dict = ps[term]["items"]
+
+    if "$ref" in search_dict.keys():
+        schema_name = search_dict["$ref"]
+    elif len(set(search_dict.keys()).intersection(["anyOf", "oneOf", "allOf"])) > 0:
+        # Get the correct jsonld relationship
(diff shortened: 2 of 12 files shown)


scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+    resolver = jss.RefResolver(base_uri=parent_schema["id"], referrer=None)
+
+    # See if "items" is in the schema entry
+    search_dict = ps[term]
+    if "items" in search_dict.keys():
+        search_dict = ps[term]["items"]
+
+    if "$ref" in search_dict.keys():
+        schema_name = search_dict["$ref"]
+    elif len(set(search_dict.keys()).intersection(["anyOf", "oneOf", "allOf"])) > 0:
```
cmadjar, line 46: I wonder whether the content of this if section could be moved into a function to make the code a bit clearer. I got lost in the different if statements when reviewing the PR, LOL.

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        logger.debug(f'{dtype} not in context {schema["id"]}')
+    else:
+        specific_context[dtype] = context[dtype]
+    json_object["@type"] = dtype
+    return json_object, specific_context
+
+
+def gen_jsonld_outpath(dats_json_f, out_path):
+    """
+    This function generates a output file path for the annotated DATS jsonld file.
```
cmadjar, line 174:
```suggestion
    This function generates an output file path for the annotated DATS JSONLD file.
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+    elif out_path.suffix != "":
+        # We have most likely gotten a path to a non-existent file, let's use that
+        return out_path
+    elif not out_path == dats_json_f.parent:
+        # We are saving this somewhere other than the original folder
+        out_name = f"{dats_json_f.parent.name}_{out_name}"
+    if out_path.is_dir():
+        dats_jsonld_f = out_path / out_name
+    elif out_path.parent.is_dir():
+        # TODO: check if there is a way to explicitly look for a file that doesn't exist
```
cmadjar, line 198: wouldn't `my_file.is_file()` or `my_file.exists()` do this?

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        logger.info(f"I will write to {out_path.resolve() = }")
+        dats_jsonld_f = out_path.parent / out_name
```
cmadjar, line 200: Does this work? I usually use the following (which avoids // in the path):
```suggestion
        dats_jsonld_f = os.path.join(out_path.parent, out_name)
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+    else:
+        raise Exception(
+            f"{out_path = } for {dats_json_f.resolve} is not a path to a file or directory. "
+            f"I don't know how to store the output"
```
cmadjar, line 204:
```suggestion
            f"I don't know where to store the output"
```

cmadjar: maybe, could hint the user to provide a valid path with option --out?

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+    return dats_jsonld_f
+
+
+def dats_to_jsonld(dats_f, schema_f, context_dir, out_path=None, clobber=False):
+    """
+    Helper function to load the inputs and store the annotated DATS.jsonld file
+    """
+    # TODO: log how many terms were not annotated because they are missing from context
```
cmadjar, line 213: that would indeed be nice to have that summary once the script is done running!

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+    tic = time.time()
+    dats_jsonld_f = gen_jsonld_outpath(dats_f, out_path)
+    if dats_jsonld_f.is_file():
+        if clobber:
+            logger.warning(
+                f"{dats_jsonld_f.resolve()} already exists and {clobber = }. "
+                f"I will overwrite the file now!"
```
cmadjar, line 220:
```suggestion
                f"The file {dats_jsonld_f.resolve()} will be overwritten now!"
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            logger.warning(
+                f"{dats_jsonld_f.resolve()} already exists and {clobber = }. "
+                f"Consider setting the clobber flag to overwrite existing files"
+            )
+            return
+
+    dats_json = json.load(open(dats_f))
+    schema = json.load(open(schema_f))
+
+    # TODO validate the DATS.json before annotation
```
cmadjar, line 232: ooooh, yes. That would be awesome!

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+
+    logging.basicConfig(format="%(levelname)s:  %(message)s", level=logging.INFO)
+    parser = argparse.ArgumentParser(
+        description="Annotate a DATS.json file to DATS.jsonld given a schema"
+    )
+    parser.add_argument(
+        "dats_path",
+        type=pal.Path,
+        help="""
+                If this is a path to a DATS file, then only this DATS file is annotated.
```
cmadjar, line 260:
```suggestion
                If this is a path to a DATS file, then only this DATS file will be annotated.
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        "--dats_schema",
+        type=pal.Path,
+        default=SCHEMA_DIR / "dataset_schema.json",
+    )
+    parser.add_argument("-dc", "--dats_context_dir", type=pal.Path, default=CONTEXT_DIR)
+    parser.add_argument(
+        "--out",
+        type=pal.Path,
+        default=None,
+        help="Where to put the output (default = in the same folder).",
```
cmadjar, line 276:
```suggestion
        help="Where to create the JSONLD file(s) (default = in the same folder).",
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        help="""
+                If this is a path to a DATS file, then only this DATS file is annotated.
+                If this is a path to a directory, then each subdirectory of this directory is expected
+                to contain a DATS file called DATS.json. Each of these files will then be annotated iteratively
+                """,
+    )
+    parser.add_argument(
+        "-ds",
+        "--dats_schema",
+        type=pal.Path,
```
cmadjar, line 268: should there be a help text here?

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            dats_f=args.dats_path,
+            schema_f=args.dats_schema,
+            context_dir=args.dats_context_dir,
+            out_path=args.out,
+            clobber=args.clobber,
+        )
+    elif args.dats_path.is_dir():
+        files_to_convert = list(args.dats_path.glob("*/DATS.json"))
+        if not args.out.is_dir():
+            logger.warning(f"I will create the output folder at {args.out.resolve()}")
```
cmadjar, line 292:
```suggestion
            logger.warning(f"The {args.out.resolve()} folder will be created and JSONLD files will be saved in it.")
```

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            args.out.mkdir()
+        if files_to_convert is None:
+            logger.error(
+                f"could not find any DATS.json files in subdirectories of {args.dats_path.resolve()}"
+            )
+        else:
```
cmadjar, line 298:
```suggestion
            )
            sys.exit()
```

cmadjar: Instead of adding a long else statement, could exit the script in the if statement above and remove the extra indentation of the else code.

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            logger.info(
+                f"We have found {len(files_to_convert)} files to convert at {args.dats_path.resolve()}"
```
cmadjar, line 300:
```suggestion
                f"Found {len(files_to_convert)} files to convert at {args.dats_path.resolve()}"
```

cmadjar (changes requested):
@surchs thank you for submitting this!! I did not test the script yet but had a few comments while reviewing the code.

surchs:
Thank you @cmadjar for your comments!

surchs:
@mathdugre: my PR adds two submodules (DATS-context and DATS-schema) that this script needs to run. I think they aren't available in the CI, and therefore some tests are failing. @zxenia suggested I talk to you for suggestions on how to make them available. I saw that her dats-validator script (that also relies on a submodule to be available) is set up in the [Circle CI config](https://github.com/CONP-PCNO/conp-dataset/blob/master/.circleci/config.yml#L51). What do you suggest we do to replicate this for the current PR?

scripts/dats_jsonld_annotator/README.md
```diff
@@ -0,0 +1,29 @@
...
+**Option 1**:
+
+Annotate all files called `DATS.json` in the subdirectories of a directory.
+
+```shell
+python annotator.py /path/to/project/folder --out /path/to/output/directory
+```
+
+The `--out` parameter is optional, if it is not specified, the files will be generated in the
+same directory as the original DATS.jsonld.
```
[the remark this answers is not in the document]
surchs: yes indeed, thank you!

scripts/dats_jsonld_annotator/annotator.py
surchs: I don't think so but I also think that this comment is outdated. My goal was to find some check whether the submitted path points to a file. The solution I came up with is to check for a non-empty, dot-separated file suffix ([in line 189](https://github.com/CONP-PCNO/conp-dataset/pull/640/files#diff-e5ad809a7eb9c3c0fa56ab26cd60aab38eb52f7af96b096b0302464988592650R189)).  

I think this section is not doing a good job. The only way to arrive here is if you supplied the path to a non-existent directory, e.g. `/all/these/dirs/exist/but_not_this_one`. Then the current code would put the results in `/all/these/dirs/exist/` which is most likely unexpected. So maybe here we should rather create the directory `but_not_this_one`. What do you think @cmadjar

scripts/dats_jsonld_annotator/annotator.py
surchs: Yes, both would work. I am using [pathlib](https://docs.python.org/3/library/pathlib.html) in this script because I find it makes for more intuitive syntax.

scripts/dats_jsonld_annotator/annotator.py
surchs: Yes, I like this suggestion. Will change. Thanks!

scripts/dats_jsonld_annotator/annotator.py
surchs: Yes indeed. Thanks for finding this! I will add a help string.

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            ref_name = ref["$ref"]
+            _schema_uri, _schema = resolver.resolve(ref_name)
+            if _schema_uri == resolver.base_uri:
+                _schema = parent_schema
+            if jss.Draft4Validator(_schema).is_valid(json_object):
+                possible_keys.append(ref_name)
+        if len(possible_keys) > 1:
+            # TODO: decide if we let the user pick which option to go with
+            logger.debug(f"I got more than one option for {term}: {possible_keys}")
+        elif len(possible_keys) == 0:
```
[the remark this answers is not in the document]
surchs: I think this could get indeed very verbose. There will be thousands of "matches". We could add it and log at the debug level if you think this would be an important information to keep around

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            # TODO: we might also want to raise here as this shouldn't happen
+            logger.warning(
+                f"I have no fitting schema for {json_object} {term} among {search_dict[schema_rel]}"
+            )
+            return None
+        # If anything fits, just pick the first one
+        # TODO: we may want to leave this up to the user here, particularly if the instances
+        #       map to different / meaningful things in SDO
```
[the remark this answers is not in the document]
surchs: What do you have in mind here? It's pretty frequent that more than one schema option would be valid. So for now I logged this as "debug" and don't show by default. If we were to raise the log level, these messages would most likely overwhelm / dominate the terminal output. But I agree, this information is relevant.

Regarding the TODO: I think letting the user choose the "correct" schema sounds like a good strategy. But the reality is that the majority of the schema @types (i.e. the schema.org classes) are sdo:Thing. So in the final jsonld output it probably rarely really makes a difference if we choose one or the other from a list of valid schemata. Given how frequently such a choice would have to be made, this would make the annotation pretty laborious / slow. So we should probably discuss this together.

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+    :param parent_schema: The DATS JSON schema that corresponds to the json_object parameter
+    :param term: the string term for which the schema shall be found
+    :param json_object:  the JSON object that contains the term and to which the parent_schema corresponds
+    :return: the DATS JSON schema for the term
+    """
+    ps = parent_schema["properties"]
+    if term not in ps.keys():
+        # TODO: Decide whether to raise an Exception here
+        # The only way this could happen on a valid dataset is if someone has specified additional properties
+        # We'd like to discourage this
```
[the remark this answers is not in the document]
surchs: Let's turn this into a warning and remove the TODO here

scripts/dats_jsonld_annotator/annotator.py
surchs: let's keep as is with the new comments and maybe revisit later if there is another solution to finding valid schemata for a given json object

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+        for ref in search_dict[schema_rel]:
+            # Get the schema
+            ref_name = ref["$ref"]
+            _schema_uri, _schema = resolver.resolve(ref_name)
+            if _schema_uri == resolver.base_uri:
+                _schema = parent_schema
+            if jss.Draft4Validator(_schema).is_valid(json_object):
+                possible_keys.append(ref_name)
+        if len(possible_keys) > 1:
+            # TODO: decide if we let the user pick which option to go with
```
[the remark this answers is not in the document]
surchs: leave as is, but create an issue to have a log file with all of this info because it would not be included in the terminal log

scripts/dats_jsonld_annotator/annotator.py
```diff
@@ -0,0 +1,323 @@
...
+            logger.debug(f"I got more than one option for {term}: {possible_keys}")
+        elif len(possible_keys) == 0:
+            # TODO: we might also want to raise here as this shouldn't happen
```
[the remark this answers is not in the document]
surchs: We also have the dats-validator. So we don't need to raise here as this is very unlikely to happen and would be excluded from the jsonld

scripts/dats_jsonld_annotator/annotator.py
surchs: let's create an extension issue

scripts/dats_jsonld_annotator/annotator.py
surchs: let's add a simple jsonschema validation (or re-use what Ksenia's validator does)

cmadjar (approved):
LGTM.

PR #660, opened 2021-06-16
[FIX] CircleCI config to allow sub-modules installation in the build phase

mathdugre:
## Description

Currently, only the `scripts/dats_validator/conp-dats` sub-modules is installed during the build phase of the tests.
#640 demonstrates the issue with this approach.

This PR proposed to recursively install the `scripts` directory during the build phases. This should avoid further failure in the future when adding more sub-modules.


cmadjar (approved):
LGTM

PR #663, opened 2021-06-23
Adding microstructure_informed_connectomics dataset

emmetaobrien:
This PR adds the `microstructure_informed_connectomics` dataset

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -166,6 +166,3 @@
 [submodule "projects/spins_travelling_phantoms"]
 	path = projects/spins_travelling_phantoms
 	url = https://github.com/conpdatasets/spins_travelling_phantoms.git
-[submodule "projects/microstructure_informed_connectomics"]
-	path = projects/microstructure_informed_connectomics
-	url = https://github.com/conpdatasets/microstructure_informed_connectomics.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -166,3 +166,6 @@
 [submodule "projects/spins_travelling_phantoms"]
 	path = projects/spins_travelling_phantoms
 	url = https://github.com/conpdatasets/spins_travelling_phantoms.git
+[submodule "projects/microstructure_informed_connectomics"]
+	path = projects/microstructure_informed_connectomics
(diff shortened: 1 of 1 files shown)


emmetaobrien:
This dataset is **not** for public release yet, as there is an issue with one of the patients involved who may need to be removed from the study;  we are awaiting confirmation on this from the data provider.

cmadjar:
Added the blocked label as the dataset is not ready to be shared publicly on the portal yet. Some data might still be modified.

cmadjar (changes requested):
The README.md file appears incomplete. There is a sentence to ask users to cite a reference but the reference is not given.

The README.md format should be updated as follows:
```
# Microstructure-Informed Connectomics (MICs) dataset

### Description

The MICs dataset provides raw and fully processed multimodal neuroimaging data acquired in 50 healthy control participants at 3T. Modalities include high-resolution anatomical (T1w), microstructurally-sensitive (T1 mapping), diffusion-weighted and resting-state functional imaging. In addition, MICs provides ready-to-use connectomes built across multiple parcellation schemes and spatial scales based on each of these modalities.

### Citation

(only if there is a citation to add though)
```

Regarding the DATS file:

- types and keywords should be all lower case as we discussed during the harmonization work (unless they are acronyms)
- interesting, there is a `é` in Québec value but a `e` in Montreal in the DATS file. Can't believe I noticed that, haha

emmetaobrien:
The data provider has just uploaded an updated version of the entire dataset, with changes in pre-processing steps that will have small global effects as well as removing the patient mentioned above.  I have removed the existing data and am processing the new version.

emmetaobrien:
Processing the new dataset is now complete.

emmetaobrien:
Changed dataset name from `microstructure_informed_connectomics` to `mica-mics` at data provider's request.

emmetaobrien:
Updated `README.md` and `DATS.json` received from data provider have been added to the `mica-mics` dataset, so this should be ready for final review and release.

cmadjar:
Thank you @emmetaobrien. Just sent the link to the two files to the com committee for final review. Once approved I will merge the PR :)

PR #670, opened 2021-07-07
[FIX] docker build workflow failed due to changes in pip installation.

mathdugre:
## Description
The workflow for building a Docker image is currently failing.
The issue is caused by the new way Ubuntu installs pip3. It nows create a symlink to pip which conflicts with the current Dockerfile.
This PR remove the manual symlink from pip3 -> pip and feeze the version of Ubuntu to avoid further  unexpected failures.


glatard (approved):
looks good to me!

PR #675, opened 2021-07-14
Fix auto-archive script

mathdugre:
## Description

This PR fixes two issues:
1. Datasets crawled from Zenodo and OSF were skipped during the archiving.
2. Datasets in sub-folder failed to be archived.

## Related issues (optional)

Fix #672

PR #682, opened 2021-07-23
[FIX] archives were stored in the wrong location.

mathdugre:
## Description

This PR fix a bug introduced in #675 where the archives are not stored in the right location.


cmadjar (approved):
LGTM. Thank you @mathdugre

PR #686, opened 2021-08-03
Add logo to the new bigbrain datasets

cmadjar:
The two new BigBrain datasets are missing the logo in their directory. This PR updates the two BigBrain dataset to their latest commit that include the logo.png image.


emmetaobrien (approved):
Seems straightforward and entirely fine.

PR #693, opened 2021-08-18
Update auto archive script with no default min size

cmadjar:
## Description

This modifies the default behaviour of the auto-archive script since we will not be restricting anymore the size of downloadable archives as discussed during last week's developer meeting.


emmetaobrien (approved):
this all makes sense to me.

PR #697, opened 2021-09-09
Updating to correct detached heads

emmetaobrien:
This PR updates the following subdatasets to the latest versions to correct detached-head errors:

```
/projects/openpain/thermal
/scripts/dats_jsonld_annotator/context
/scripts/dats_validator/conp-dats
```

PR #698, opened 2021-09-09
Test to see if that fixes the test failure reported in 697

cmadjar:
## Description

In PR #697, several datasets that were crawled by OSF or Zenodo failed the tests after updating the conp-dats submodule. This PR should be able to fix the issues flagged by the tests for the retrospective OSF datasets that do not have a distribution->formats field.

Suggestion was to completely skip the DATS tests for those retrospective datasets until we get a chance to contact the authors of the datasets so that their DATS files can be fixed.

### DIFF
--- a/tests/template.py
+++ b/tests/template.py
@@ -48,7 +48,7 @@ class Template:
             )
 
     def test_has_valid_dats(self, dataset):
-        
+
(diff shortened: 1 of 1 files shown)


emmetaobrien (approved):
This looks reasonable to me.

PR #700, opened 2021-09-27
Uninstall a dataset once the archive has been created

cmadjar:
## Description

Because some datasets to archive are actually quite large, disk space gets quickly filled up. Once an archive is created, the auto-archive script will uninstall the DataLad dataset that has been archive to manage space better on the server.

Note: initially tried with `datalad drop` but realized quickly that it did not really free up space on the server so instead, we have to run `datalad uninstall`. See https://github.com/datalad/datalad/issues/6009 created on the DataLad GitHub repo.

### DIFF
--- a/scripts/auto_archive.py
+++ b/scripts/auto_archive.py
@@ -228,6 +228,9 @@ if __name__ == "__main__":
                         archive_name=archive_name,
                         version=version,
                     )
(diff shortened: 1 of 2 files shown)


emmetaobrien (approved):
lgtm

Issue #701, opened 2021-10-08
If derivedFrom field is populated in DATS file, automatically add the raw dataset to the datalad repo


cmadjar:
## Purpose

When a user is uploading a new dataset that was derived from another CONP dataset, the derivedFrom field of the DATS file should be populated.

Suggestion to read that field when present and automatically add the DataLad raw dataset as a submodule of the generated dataset.

Note: should be done in the base class as pertinent for both OSF and Zenodo crawler.

PR #704, opened 2021-10-25
If derivedFrom field is populated in DATS file, automatically add the raw dataset to the datalad repo

cmadjar:
## Description

This modifies the base crawler pipeline to read the DATS.json file of the crawled dataset and check whether there is an 'derivedFrom' entry in the extra properties. If there is one, then add the source dataset as a submodule to the crawled dataset.

## Related issues 

#701

### DIFF
--- a/scripts/Crawlers/BaseCrawler.py
+++ b/scripts/Crawlers/BaseCrawler.py
@@ -279,6 +279,7 @@ class BaseCrawler:
                         print(f"Found existing DATS.json at {existing_dats_path}")
                     if existing_dats_path != dats_path:
                         os.rename(existing_dats_path, dats_path)
+                    self._add_source_data_submodule_if_derivedfrom_conp_dataset(dats_path)
                 else:
                     self._create_new_dats(
                         dataset_dir,
@@ -552,3 +553,20 @@ Functional checks:
                         return os.path.join(file_path, subfile_name)
             elif file_name.lower() == "dats.json":
                 return file_path
+
+    def _add_source_data_submodule_if_derivedfrom_conp_dataset(self, dats_json):
+        f = open(dats_json, 'r')
+        metadata = json.loads(f.read())
+        f.close()
+
+        source_dataset_link = None
+        source_dataset_id = None
+        for property in metadata['extraProperties']:
+            if property['category'] == 'derivedFrom':
+                source_dataset_link = property['values'][0]['value']
+            if property['category'] == 'parent_dataset_id':
+                source_dataset_id = property['values'][0]['value']
+
+        if 'github.com' in source_dataset_link:
+            d = self.datalad.Dataset(os.path.join(dataset_dir, source_dataset_id))
+            d.create()
--- a/scripts/Crawlers/BaseCrawler.py
+++ b/scripts/Crawlers/BaseCrawler.py
@@ -559,7 +559,9 @@ Functional checks:
             elif file_name.lower() == "dats.json":
                 return file_path
 
-    def _add_source_data_submodule_if_derivedfrom_conp_dataset(self, dats_json, dataset_dir):
+    def _add_source_data_submodule_if_derivedfrom_conp_dataset(
+            self, dats_json, dataset_dir
+    ):
         f = open(dats_json)
         metadata = json.loads(f.read())
         f.close()
--- a/scripts/Crawlers/BaseCrawler.py
+++ b/scripts/Crawlers/BaseCrawler.py
@@ -280,7 +280,7 @@ class BaseCrawler:
                     if existing_dats_path != dats_path:
                         os.rename(existing_dats_path, dats_path)
                     self._add_source_data_submodule_if_derivedfrom_conp_dataset(
-                        dats_path
+                        dats_path, dataset_dir
                     )
                 else:
                     self._create_new_dats(
@@ -323,7 +323,7 @@ class BaseCrawler:
                         if existing_dats_path != dats_path:
                             os.rename(existing_dats_path, dats_path)
                         self._add_source_data_submodule_if_derivedfrom_conp_dataset(
-                            dats_path
+                            dats_path, dataset_dir
                         )
                     else:
                         self._create_new_dats(
@@ -559,7 +559,7 @@ Functional checks:
             elif file_name.lower() == "dats.json":
                 return file_path
 
-    def _add_source_data_submodule_if_derivedfrom_conp_dataset(self, dats_json):
+    def _add_source_data_submodule_if_derivedfrom_conp_dataset(self, dats_json, dataset_dir):
         f = open(dats_json)
         metadata = json.loads(f.read())
         f.close()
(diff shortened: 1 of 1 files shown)


cmadjar:
Passing the tests now. @surchs would you happen to have some time to review this PR by any chance? Let me know if not and I'll try to find another victim ;).

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,22 @@ def _check_dats_present(self, directory):
                         return os.path.join(file_path, subfile_name)
             elif file_name.lower() == "dats.json":
                 return file_path
+
+    def _add_source_data_submodule_if_derivedfrom_conp_dataset(
```
surchs, line 562: for consistency, I'd make this fully snake-case
```suggestion
    def _add_source_data_submodule_if_derived_from_conp_dataset(
```
and then refactor above

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,22 @@ def _check_dats_present(self, directory):
...
+        self, dats_json, dataset_dir
+    ):
+        f = open(dats_json)
```
surchs, line 565: I would suggest using the `with` context here. i.e. `with open(..., 'r') as f: ... `

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,22 @@ def _check_dats_present(self, directory):
...
+        metadata = json.loads(f.read())
+        f.close()
+
+        source_dataset_link = None
+        source_dataset_id = None
+        for property in metadata["extraProperties"]:
```
surchs, line 571: I don't think `extraProperties` is part of the required DATS properties. Could there be a case where the `"extraProperties"` key does not exist? I think it'd make sense to check this here.

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,22 @@ def _check_dats_present(self, directory):
...
+            if property["category"] == "derivedFrom":
+                source_dataset_link = property["values"][0]["value"]
+            if property["category"] == "parent_dataset_id":
+                source_dataset_id = property["values"][0]["value"]
+
+        if "github.com" in source_dataset_link:
```
surchs, line 577: I think this will break if `source_dataset_link` is not iterable, e.g. if it is `None`. This might happen if no `property["category"] == "derivedFrom"`. You could make that the first condition, i.e. `if source_dataset_link is not None and ...`.

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,22 @@ def _check_dats_present(self, directory):
...
+    ):
+        f = open(dats_json)
+        metadata = json.loads(f.read())
+        f.close()
+
+        source_dataset_link = None
+        source_dataset_id = None
+        for property in metadata["extraProperties"]:
+            if property["category"] == "derivedFrom":
+                source_dataset_link = property["values"][0]["value"]
```
surchs, line 573: Is there some guarantee that all of these values are going to be present? The [DATS dataset schema](https://github.com/CONP-PCNO/schema/blob/master/dataset_schema.json#L202) does not require any fields for [`category_values_pair_schema`](https://github.com/CONP-PCNO/schema/blob/master/category_values_pair_schema.json) or the [`annotation_schema`](https://github.com/CONP-PCNO/schema/blob/master/annotation_schema.json) that will be nested in it. So unless these are forced by the crawler somewhere else, this line could raise `KeyError` or `IndexError` unexpectedly here.

surchs (changes requested):
Hi @cmadjar, thanks for the PR! I had some question about the code because I am not sure where the DATS file comes from that is being parsed here. Or rather: how much we know about the DATS file before we parse it.

scripts/Crawlers/BaseCrawler.py
cmadjar: fixed in the next commit

scripts/Crawlers/BaseCrawler.py
cmadjar: thank you! Fixed in the next commit

scripts/Crawlers/BaseCrawler.py
cmadjar: good catch. Done in the next commit

scripts/Crawlers/BaseCrawler.py
cmadjar: I added an if statement right above to check if the property exists. If it does not exist, then it will simply return from the function as there is no information regarding potential source data.

scripts/Crawlers/BaseCrawler.py
cmadjar: I just added a few if statements above to check whether each values are present, if not, then go on to the next property. Not sure if there is a more elegant way to check those 3 things...

cmadjar (commented):
@surchs Thank you for your review!

The DATS file here corresponds to the DATS.json file a user will submit to Zenodo (or OSF) with their datasets. Instead of automatically creating it, the pipeline downloads it from the Zenodo or OSF API. Then, here, we read the file in order to determine whether the dataset is a derived dataset or not based on the values stored in the `derivedFrom` extra property.

See responses to your comments below. Let me know if there should be other improvements. 

Thank you so much for your help!! :-D

scripts/Crawlers/BaseCrawler.py
surchs: I think this is mostly a style question but you could make this a bit cleaner by using `try` and `except`:

```
try:
    source_dataset_link = property["values"][0]["value"]
except (KeyError, IndexError):
    continue
```

scripts/Crawlers/BaseCrawler.py
```diff
@@ -552,3 +558,29 @@ def _check_dats_present(self, directory):
...
+            if type(property["values"]) != list:
+                continue
+            if "value" not in property["values"][0].keys():
+                continue
+            if property["category"] == "derivedFrom":
+                source_dataset_link = property["values"][0]["value"]
+            if property["category"] == "parent_dataset_id":
+                source_dataset_id = property["values"][0]["value"]
+
+        if source_dataset_link and "github.com" in source_dataset_link:
```
surchs, line 584: I would be explicit here and check that `source_dataset_link is not None`. `None` is "falsy" and here the variable is declared just above, but it's still better to [be consistent](https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations).

cmadjar: 👍  Done in the last commit

scripts/Crawlers/BaseCrawler.py
cmadjar: Niice!! Done in the next commit. Thank you!

cmadjar:
@surchs implemented your suggestions. Thank you! Let me know if you see something else that needs improving :)

surchs (approved):
lgtm @cmadjar :tada:

PR #708, opened 2021-11-04
Update projects/1000GenomesProject to most recent version

emmetaobrien:
Updates 1000GenomesProject for https://github.com/conpdatasets/1000GenomesProject/pull/14

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -22,7 +22,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
       - name: Build and push
         id: docker_build
-        uses: docker/build-push-action@v2.6.1
+        uses: docker/build-push-action@v2.7.0
         with:
           push: true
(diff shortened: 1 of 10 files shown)


cmadjar (commented):
@emmetaobrien why is there so many files changed? I was expecting to see only the 1000GenomesProject

emmetaobrien:
That was an error, those are other changes from other not yet accepted PRs. I will try to clean that up this afternoon.

cmadjar:
@emmetaobrien I still see changes to unrelated submodules in this PR. There should only be updates to the 1000GenomesProject submodule in that PR. Thanks!

surchs:
@emmetaobrien: the fails here are a result of the updated context submodules. There is a context fixture in the annotator test that  needs to be updated for the extended DATS contexts. This should however happen in a separate PR, I'll open an issue and fix this!

I agree with @cmadjar here: roll-back all changes that are not related to `Update projects/1000GenomesProject to most recent version` and then let's deal with the rest separately. That will also resolve the failing test for this PR.

cmadjar (approved):

PR #711, opened 2021-11-10
Modifies the validator to check that the species URL is in identifier instead of identifierSource

cmadjar:
## Description

As discussed during the CONP dev call today, this modifies the DATS validator so that it flags all the datasets that need to be modified where the species URI is under `identifierSource` instead of the `identifier` key.

This will flag for us which datasets need to be changed. 

Related to https://github.com/CONP-PCNO/schema/issues/8


emmetaobrien (approved):
Looks good to me.

PR #713, opened 2021-11-11
Update the DATS context submodule

surchs:
The CONP DATS schema is used as a submodule for the DATS annotator. This PR updates the submodule in order to include the recent schema extensions. A test fixture also had to be updated to match the updated schema.

This came up in #708


emmetaobrien (approved):
All looks reasonable to me.

PR #770, opened 2022-04-28
Update bigbrain to add ARK ID

cmadjar:
## Description

This updates the BigBrain dataset to add the ARK ID to the DATS file now that it has been generated by the portal.

### DIFF
--- a/tests/template.py
+++ b/tests/template.py
@@ -162,12 +162,13 @@ class Template:
             input_submodule_commit = input_submodule_repo.commit()
             input_submodule_url = input_submodule.url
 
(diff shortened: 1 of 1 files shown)


cmadjar:
added the blocked and tests labelled as this is initially to test whether the tests will pick up that bigbrain-datalad has not been updated for the derived datasets.

It is also blocked by https://github.com/conpdatasets/bigbrain-datalad/pull/13 while we are thinking of the best way to add the ARK ID information to the DATS schema.

emmetaobrien (approved):
LGTM

PR #771, opened 2022-05-03
Fix provider issue for CHBMP - config file was not properly named

cmadjar:
The provider config file was not properly named when added to the dataset (loris-cfg) resulting in errors when trying to download the dataset. This PR pulls the latest changes of CHBMP that renamed the file correctly (loris.cfg)


emmetaobrien (approved):
looks like it should work.

PR #778, opened 2022-05-20
Add initial CCNA_COMPASS-ND data descriptors

emmetaobrien:
This PR adds the initial CCNA_COMPASS-ND dataset, containing a `README.md`, a `DATS.json` and a `logo.png`.


cmadjar (changes requested):
@emmetaobrien I think the test failure is related to the fact that there is no .datalad directory. Even though there is no data yet to accompany the descriptions, it might be good to still have the repo set up as a datalad repo.

PR #791, opened 2022-06-09
add C-OPN dataset

cmadjar:
## Description

This pull request adds descriptors for the C-OPN dataset, replacing https://github.com/CONP-PCNO/conp-dataset/pull/787.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -217,3 +217,6 @@
 [submodule "projects/braincode_POND_Registry_Imaging_Data_Release"]
 	path = projects/braincode_POND_Registry_Imaging_Data_Release
 	url = https://github.com/conpdatasets/braincode_POND_Registry_Imaging_Data_Release
+[submodule "projects/C-OPN_LORIS"]
+	path = projects/C-OPN_LORIS
+	url = https://github.com/conpdatasets/C-OPN_LORIS.git


.gitmodules
```diff
@@ -217,3 +217,6 @@
 [submodule "projects/braincode_POND_Registry_Imaging_Data_Release"]
 	path = projects/braincode_POND_Registry_Imaging_Data_Release
 	url = https://github.com/conpdatasets/braincode_POND_Registry_Imaging_Data_Release
+[submodule "projects/C-OPN_LORIS"]
+	path = projects/C-OPN_LORIS
+	url = https://github.com/conpdatasets/C-OPN_LORIS.git
```
GHPBZ, line 222: This description is fine and about 2/3 of was used for the description field in the DATS.json.  We need to check in with Samir and JB for a realistic timeline of data release here so we know if we have anything to point to.  Depending on that timeline, it may also be a candidate for a 'coming soon' banner if we implement that feature.

emmetaobrien:
still under review

emmetaobrien:
Not confirmed defunct yet.

GHPBZ:
Alright ... Let's put this to bed.  After a failed attempt earlier in 2024, there is a reasonable chance at getting an new open-access Parkinson's release together in the next few months, but we'll just create a new issue when that time arises.

PR #807, opened 2023-01-12
Adding two new BrainCODE datasets

emmetaobrien:
This PR adds two new Brain-CODE datasets, EpLink and CP-NET.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -220,3 +220,9 @@
 [submodule "projects/braincode_Epilepsy_Priority_Setting_Partnership"]
 	path = projects/braincode_Epilepsy_Priority_Setting_Partnership
 	url = https://github.com/conpdatasets/braincode_Epilepsy_Priority_Setting_Partnership
+[submodule "projects/braincode_CP-NET"]
+	path = projects/braincode_CP-NET
+	url = https://github.com/CONP-PCNO/braincode_CP-NET
+[submodule "projects/braincode_EpLink"]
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -220,3 +220,9 @@
 [submodule "projects/braincode_Epilepsy_Priority_Setting_Partnership"]
 	path = projects/braincode_Epilepsy_Priority_Setting_Partnership
 	url = https://github.com/conpdatasets/braincode_Epilepsy_Priority_Setting_Partnership
+[submodule "projects/braincode_CP-NET"]
+	path = projects/braincode_CP-NET
+	url = https://github.com/CONP-PCNO/braincode_CP-NET
```
GHPBZ, line 225: This README looks good and the overview is the same as the description field in the DATS.json.

.gitmodules
```diff
@@ -220,3 +220,9 @@
...
+[submodule "projects/braincode_EpLink"]
+	path = projects/braincode_EpLink
+	url = https://github.com/CONP-PCNO/braincode_EpLink
```
GHPBZ, line 228: Same as for CP-NET: the README looks good and the overview is the same as the description field in the DATS.json.

GHPBZ (approved):
I assume both the EpLink and CP-NET DATS.json files have passed the DATS validator?  In this case, I think we can move ahead with these and notify Brendan when we are done.

PR #810, opened 2023-01-19
temporary workaround for validator issue.

emmetaobrien:
Some of the braincode datasets have been failing validation, because the validator expects to to find login credentials associated with restricted or private data, and though these datasets were marked as restricted in the `access`->`authorization` field in `DATS.json`, they contain only metadata and have no credentials.  The validator software needs to be fixed to address this issue, but for a short-term fix, the relevant field was adjusted to `public`.


tkkuehn (approved):
LGTM with the caveat that if I can quickly get the tests resolved without merging this we can close it.

tkkuehn:
Side note: It seems that I don't have the status in this repo/organization to have my approving review unblock the PR.

emmetaobrien:
> Side note: It seems that I don't have the status in this repo/organization to have my approving review unblock the PR.

I have just given you write access, which looks to have addressed that issue.

emmetaobrien:
Rendered obsolete by #811.

PR #811, opened 2023-01-26
Update auth test to handle no annexed content

tkkuehn:
## Description

This updates the authentication function used in testing to check whether there is any annexed content, and ignores the lack of auth info if there is not.

### DIFF
--- a/tests/functions.py
+++ b/tests/functions.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager
 from functools import reduce
 
 import datalad.api as api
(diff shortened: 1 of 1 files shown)


tkkuehn:
Note that this should allow us to close #810 without merging it.

emmetaobrien:
All the test failures with this are known download errors on specific datasets, it accurately corrects behaviour on the braincode data, so I think it looks good to go.

emmetaobrien (approved):

PR #813, opened 2023-01-26
Add braincode_EpLink and braincode_CP_NET datasets

emmetaobrien:
This PR replaces #807, adding two new braincode datasets, `braincode_EpLink` and `braincode_CP_NET`.

PR #814, opened 2023-01-26
correcting error in #813

emmetaobrien:
This PR corrects an error in #813  where the new datasets were added at the wrong level in the directory structure.


tkkuehn (approved):
Sorry I didn't catch this the first time, looks good

tkkuehn (commented):
Temporarily retracting my approval to look into what went wrong with the test

tkkuehn (approved):
Everything looks like it's in order now!

PR #818, opened 2023-01-31
Fix pre commit

tkkuehn:
## Description

The "pre-commit hooks" GitHub action fails because the flake8 pre-commit hook is misspecified. This fixes that issue and incorporates the results of `pre-commit run --all-files`, and a minor change to the json-ld annotator, making the exception raised slightly more specific to satisfy flake8.


emmetaobrien (approved):
Looks logical to me.

Issue #782, opened 2022-06-03
Minor format issue with validation of git configuration files.


emmetaobrien:
The style guidelines that our validation process currently uses in tests require an empty line at the end of each file.  This causes potential issues with `datalad` and `git-annex` related configuration files in our datasets, such as `.gitmodules` and `.git/config`.  The process of adding a new dataset via `datalad` concatenates a couple of new lines to the end of these files after the empty line, and this has the potential for causing problems with the underlying `git` commands reading that empty line as a premature EOF.  In the short term this is being addressed by manually editing the config files when new datasets are added, but it would be more efficient to exempt the relevant configuration files from this check.

emmetaobrien:
Reopening this to confirm that https://github.com/CONP-PCNO/conp-dataset/pull/819 appears to have fixed it, so that it doesn't just look closed because it expired.

PR #819, opened 2023-02-16
Allow missing newline at EOF for git config files

tkkuehn:
## Description

This is a first attempt at solving #782. The pre-commit config is the only place I could find that checks for a newline at EOF, and I couldn't find a mechanism for the pre-commit to apply to subdatasets, so I'm not really confident that this will have the intended effect for subdatasets.

I have verified that it works for the root dataset, and if someone could point me at an instance of this issue causing problems, I could double check that this will work in that case.

## Checklist

Mandatory files and elements:

Functional checks:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,7 @@ repos:
       - id: check-yaml
       - id: debug-statements
       - id: end-of-file-fixer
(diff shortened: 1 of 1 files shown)


emmetaobrien (approved):
This looks like it should do exactly what we want.  Thank you.

PR #821, opened 2023-04-12
Update refseq_Mus_musculus dataset

emmetaobrien:
This pull request updates the `refseq_Mus_musculus` dataset for updates at the source.  It also adds a script for crawling the contents of `DATS.json` files in all CONP datasets and outputting them into a suitably reorganised `.csv` file to the `scripts` directory.


tkkuehn (approved):
Given what I know about DATS2csv (essentially it does what it's supposed to right now and isn't likely to be used again; it's just here for archival purposes), I haven't taken much of a look at it -- this otherwise looks fine to me.

PR #825, opened 2023-04-24
Add updated copy of Taowu dataset hosted at sftp.conp.ca

emmetaobrien:
This PR replaces the previous `Taowu` dataset, which had been giving download failures, with `Taowu_local`, the same content hosted on `sftp.conp.ca`.  It also adds an updated version of the crawling script `sftp-crawler.pl` to the `scripts` directory.

### DIFF
--- a/scripts/sftp-crawler.pl
+++ b/scripts/sftp-crawler.pl
@@ -52,5 +52,3 @@ while ($inline = <IN_LARGE>) {
 close IN_LARGE;
 
 exit();
-
-
--- a/scripts/sftp-crawler.pl
+++ b/scripts/sftp-crawler.pl
@@ -1,0 +1,56 @@
+#!/usr/bin/perl
+#
+#  sftp_crawler.pl - crawl a dataset on sftp-conp.acelab.ca
+#                    - EOB - Mar 03 2020 - reworked Apr 21 2022
+#
+# usage: sftp_crawler.pl $local_projectname $username $password $local_directory $remote_directory $inlist
+#
+#    see https://github.com/CONP-PCNO/conp-documentation/blob/master/Developers-Notes/CONP_VM_setups_and_maintenance/SFTP_community_server/CONP_data_upload_admin.md for details
+#
+#
+########################################
+
+# use lib '/home/eaobrien/perl5/lib/perl5/x86_64-linux-gnu-thread-multi';
+use Net::SFTP::Foreign;
+use Net::SFTP::Foreign::Attributes;
+use Net::SFTP::Foreign::Constants qw(:error :status);
+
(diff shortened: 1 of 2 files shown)


scripts/sftp-crawler.pl
```diff
@@ -0,0 +1,56 @@
...
+	}
+	else {
+		print "$local_file already exists\n";
+	}
+}
+close IN_LARGE;
+
+exit();
+
+
```
tkkuehn, line 56: Deleting these lines would fix the quality check failure.

tkkuehn (changes requested):
One comment on the script (which I otherwise didn't look at closely because I don't know much perl), and a couple of higher-level comments.

- https://github.com/conpdatasets/Taowu_local has no `git-annex` branch, so I can't `datalad get` any of the files.
- Why are we changing the subdataset name (which will change the ID, URL, etc. of the dataset on the portal) when the only real difference (as far as I know) is the URL of the files? Is there an advantage of this over updating the URLs at  https://github.com/conpdatasets/Taowu?

scripts/sftp-crawler.pl
emmetaobrien: These fixes are implemented in the most recent update.

tkkuehn (changes requested):
A couple more issues with the updated dataset:

- There seem to be two copies of each subject directory, one at the dataset root and one under the `taowu` subdirectory
- None of the annexed files seem to be accessible (doesn't look to me like the updated links made it to GitHub?): `git annex whereis` yields something like the following for each:
```
$ git annex whereis sub-control032057/anat/sub-control032057_T1w.nii.gz 
whereis sub-control032057/anat/sub-control032057_T1w.nii.gz (2 copies) 
  	428e1164-7df4-4c00-9757-4baa66a357f6 -- mohanad@mohanad-Aspire-TC-885:~/Desktop/taowu
   	bb7f3e7c-aaaa-40e9-a1b9-35093edd9dcb -- taowu-osf-annex-storage

```

emmetaobrien:
That looks weird in ways that make no sense with what I was seeing yesterday, I am going to recrawl the dataset, test again, and resubmit probably in a couple of hours.

emmetaobrien:
I have removed the copy of the links that did not fit with the original structure of the dataset as submitted, and explicitly submitting a PR on the `git-annex` branch in addition to the `master` branch when transferring my updated version of the Taowu dataset from my own github to `conpdatasets` seems to have fixed access to the annexed files; that last is not something I can recall needing to do separately before.

tkkuehn (approved):
This looks good to me now and works on my machine, so I'll go ahead and merge it

PR #860, opened 2023-12-01
Fix submodule path

laemtl:
Manually fix the submodule path. Will require further investigations.


emmetaobrien (approved):
Looks fine to me.

PR #924, opened 2024-08-26
open-pain test fix

laemtl:
Remove a suspicious line as part of an investigation to fix a problematic test.

### DIFF
--- a/tests/test_minimal_tests.py
+++ b/tests/test_minimal_tests.py
@@ -35,7 +35,6 @@ def test_modify_single_project(pr_files):
     "pr_files",
     [
         ("projects/preventad-open", "projects/PERFORM_Dataset__one_control_subject"),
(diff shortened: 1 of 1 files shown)


emmetaobrien (approved):

emmetaobrien:
Accepting and merging this change, because it corrected the CircleCI test failure it was intended to fix.

For reference, that problem was: the datasets used for the `modify_multi_project` test in `conp_dataset/tests/test_minimal_tests.py` are hard-coded into the program, and one of them was an openpain dataset.  This has caused persistent CircleCI test failures on PRs between  https://github.com/CONP-PCNO/conp-dataset/pull/917, which removed all the openpain datasets from `conp-dataset`, and the current PR, which had nothing to do with the contents of those PRs and may have been masking other problems; investigation is ongoing.

PR #929, opened 2024-09-05
manually re-adding Semantic Memory Encoding Task dataset.

emmetaobrien:
This PR adds https://github.com/conp-bot/conp-dataset-Semantic-Encoding-Memory-Task back into conp-dataset, without causing the download errors seen when it was added automatically.


GHPBZ (approved):

emmetaobrien:
Cancelled because this is a duplicate of an experiment that has already been manually loaded into the CONP server, identified a second time when the data provider later uploaded it to OSF.

PR #954, opened 2024-10-17
Manually adding MICA_PNI dataset back

emmetaobrien:
This PR manually adds the MICA-PNI-Precision-NeuroImaging-and-Connectomics dataset as a submodule, without the error seen in the version added by the crawler.


emmetaobrien:
Closing because the DATS.json is missing `formats` information.

PR #979, opened 2025-01-22
Adding three new eegnet datasets, and one more update.

emmetaobrien:
This PR adds the following datasets:

`eegnet_mwa`
`eegnet_ssa`
`eegnet_ssm`

and also incorporates some updates to `eegnet_tse`.


emmetaobrien:
Removing `eegnet_mwa` because there are still format errors in the DAT.json that need input from the data provider, and I would like to get the other three datasets loaded.

GHPBZ (approved):

PR #982, opened 2025-01-31
Update README.md

GHPBZ:
Minor change concerning recommended git-annex versions.


emmetaobrien (approved):
test failures due to known problems not to do with this change.

PR #1007, opened 2025-04-25
Fix LandingPage validation

laemtl:
Removing the validation on LandingPage, making it optional.


emmetaobrien (approved):
Test failures are expected, relating to inability to download randomly selected test files from registered datasets.

PR #1040, opened 2025-08-28
Adding Algonauts dataset

emmetaobrien:
This PR adds the dataset for Algonauts competitors from the CNeuromod project.


GHPBZ (commented):
For the Algonauts dataset, instead of "algonauts_2025_competitors", can we use "algonauts_project_2025_challenge" to reflect the official name of the competition?

GHPBZ (approved):

PR #1121, opened 2026-08-24
Bump 15 dataset submodules to pick up the evidence_publication DATS property

GHPBZ:
Advances the submodule pointers for 15 datasets to the commits that carry the
new `evidence_publication` property in their `DATS.json`.

No file content changes here — only submodule SHAs.

## Why

The CONP Portal links each dataset card to the NeuroLibre/Evidence publication
the data belong to (CONP-PCNO/conp-portal#845). It reads that link from DATS.

`ZenodoCrawler` was taught to emit the property in #1120, but
`update_if_necessary()` returns early when a record's Zenodo version is
unchanged, so already-crawled datasets are never regenerated. These 15 Zenodo
records are long published and will not be re-versioned, so the property was
backfilled directly into each dataset repository. This PR is the last step:
pointing the corpus at those commits.

## Verification

The bump script refuses to run unless **all 15** upstream datasets already carry
the property, so the corpus cannot end up half-updated. Each pointer is moved
only after confirming `"evidence_publication"` is present in the target commit's
`DATS.json`.

## After merging

Rebuild the Portal search index so the new field is indexed:

```bash
flask index_datasets
```

Fifteen dataset cards then show the Evidence publication strip, each with its
publication-type badge.


emmetaobrien (approved):
test failure due to known issue, not caused by this update.

